PR #26, opened 2017-09-06
Pbi119303/default groupsource

bsmithvt:


bsmithvt:
@kpaul I defaulted to null in FabricModule constructor so all subclass constructors would not have to be modified. Null is valid in this case, but you bring up a good point. Line 96 of GroupsModule.cs needs to have a null check on the DefaultPropertySettings object. Thanks.

kpaul:
@bsmithvt cool, thanks for the info.

PR #27, opened 2017-09-07
1) added swagger documentation for custom group / group->user mapping…

bsmithvt:
… feature, 2) simplified body payloads required for adding/removing roles from a group

### DIFF
--- a/Fabric.Authorization.API/Models/GroupRoleRequest.cs
+++ b/Fabric.Authorization.API/Models/GroupRoleRequest.cs
@@ -1,8 +1,17 @@
-﻿namespace Fabric.Authorization.API.Models
+﻿using System;
+
+namespace Fabric.Authorization.API.Models
 {
     public class GroupRoleRequest
     {
         public string GroupName { get; set; }
+
+        public Guid? RoleId { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


bsmithvt:
@michaelvidal24 The client does not have to specify this in the body. It will bind the GroupName path parameter to the GroupName property on the GroupRoleRequest, and it will bind the Id/RoleId properties specified in the body to the appropriate property on GroupRoleRequest. Prior to this change, the Swagger documentation indicated that an entire RoleApiModel as what is required to map a role to a group, when all that is required is a RoleId. This was confusing when I looked at the documentation, so to make this simpler (and consistent with the group->user implementation), I just eliminated swapped out the use of the RoleApiModel for GroupRoleRequest.

bsmithvt:
@michaelvidal24 To clarify, the Swagger documentation indicated a RoleApiModel was needed. I could have changed just that and not changed the actual implementation. As long as the user passed in an object that could be deserialized into a RoleApiModel, it should have been ok.

michaelvidal24:
@bsmithvt - thanks for clarifying, your changes now make sense. I forgot that Nancy will by default bind to querystring, url, and body parameters.

PR #28, opened 2017-09-11
User/bsmithvt/auth sec docs

bsmithvt:


Fabric.Authorization.API/Bootstrapper.cs
```diff
@@ -60,6 +62,13 @@ protected override void ApplicationStartup(TinyIoCContainer container, IPipeline
             SwaggerMetadataProvider.SetInfo("Fabric Authorization API", "v1",
                 "Fabric.Authorization contains a set of APIs that allow client applications to manage roles and permissions for users.");
 
+            var securitySchemeBuilder = new Oauth2SecuritySchemeBuilder();
```
michaelvidal24: Let's change the references from Auth0 to Fabric.Identity as that is our OpenID connect provider. We can use localhost:5001 as the authorization URL as a placeholder since for every installation it will likely be different.  We should also change the scope from openid to the scopes that Fabric.Authorization uses (fabric/authorization.read, fabric/authorization.write, fabric/authorization.manageclients)

PR #29, opened 2017-09-12
User/bsmithvt/user search api

bsmithvt:


Fabric.Authorization.API/Models/External/Identity/IdentityUserSearchResponse.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace Fabric.Authorization.API.Models.External.Identity
+{
+    public class IdentityUserSearchResponse
```
michaelvidal24: I'm wondering if this class would be better in a different namespace, it might be confusing to future developers that this model is really a model of a response from another service since it is located in the API/Models overall hierarchy.

Fabric.Authorization.API/Services/IdentitySearchService.cs
```diff
@@ -1,23 +1,110 @@
...
-        public async Task<IdentitySearchResponse> Search(IdentitySearchRequest request)
+        public async Task<IEnumerable<IdentitySearchResponse>> Search(IdentitySearchRequest request)
         {
+            var searchResults = new List<IdentitySearchResponse>();
             var client = await _clientService.GetClient(request.ClientId);
+            var roles = await _roleService.GetRoles();
+            var clientRoles = new List<Role>();
 
-            return new IdentitySearchResponse();
+            foreach (var role in roles)
```
michaelvidal24: Should we encapsulate this code in a function in the RoleService? Seems like we might want to query all roles for a client for another reason at some point.

Fabric.Authorization.API/Services/IdentitySearchService.cs
```diff
@@ -1,23 +1,110 @@
...
+            {
+                if (_clientService.DoesClientOwnItem(client.TopLevelSecurableItem, role.Grain, role.SecurableItem))
+                {
+                    clientRoles.Add(role);
+                }
+            }
+
+            // get all groups tied to clientRoles
+            var groups = await _groupService.GetAllGroups();
```
michaelvidal24: Roles has a collection of group ids on it, it might be faster to pull out all the group ids and query by group id rather than getting all the groups in the database.

Fabric.Authorization.API/Models/External/Identity/IdentityUserSearchResponse.cs
bsmithvt: @michaelvidal24, do you think we should move code related to external services into an entirely different project?

michaelvidal24: @bsmithvt - that's not a bad idea, but wee could also just have a namespace/folder:
>Fabric.Authorization.API.RemoteServices

and each remote service could have its own namespace:
>Fabric.Authorization.API.RemoteServices.Identity

and have a models namespace within each.

PR #30, opened 2017-09-19
User/bsmithvt/user search integration tests

bsmithvt:


Fabric.Authorization.API/Modules/IdentitySearchModule.cs
```diff
@@ -22,7 +24,7 @@ public class IdentitySearchModule : SearchModule<IdentitySearchRequest>
         {
             _identitySearchService = identitySearchService;
 
-            Get("/", async _ => await SearchIdentities().ConfigureAwait(false), null, "GetGroup");
+            Get("/", async _ => await SearchIdentities().ConfigureAwait(false), null, "SearchIdentities");
         }
 
         private async Task<dynamic> SearchIdentities()
```
michaelvidal24: We probably talked about this before, I don't recall at this point, but should we be mixing identity based scopes in authorization? I can see the potential need to secure the call to Fabric.Identity, but perhaps the scope that the client application uses is an authorization based scope that is different from fabric/authorization.read.

kpaul: I agree. I was thinking that authorization would use the identity.read scope to get the users from identity, but the client calling authorization would use either authorization.read or a new authorization scope

Fabric.Authorization.API/Services/IdentitySearchService.cs
```diff
@@ -39,21 +39,29 @@ public class IdentitySearchService
             }
 
             var client = await _clientService.GetClient(request.ClientId);
-            var clientRoles = await _roleService.GetRoles(client);
+            var clientRoles = _roleService.GetRoles(client).Result.ToList();
```
kpaul: was there an issue with using await here?

bsmithvt: @kpaul No, I just needed to ToList it to prevent the "possible multiple enumerations" warning.

michaelvidal24: @bsmithvt - can you ToList() clientRoles later? Removing the await keyword and adding the .Result will cause this function to run synchronously I believe, and may introduce deadlock issues.

Fabric.Authorization.IntegrationTests/IntegrationTestsFixture.cs
```diff
@@ -37,9 +37,9 @@ protected IDocumentDbService DbService()
             {
                 ICouchDbSettings config = new CouchDbSettings()
                 {
-                    DatabaseName = "integration-" + DateTime.UtcNow.Ticks.ToString(),
-                    Username = "",
-                    Password = "",
+                    DatabaseName = "integration-" + DateTime.UtcNow.Ticks,
```
kpaul: dont forget to revert this

PR #31, opened 2017-09-19
added handler for unhandled exceptions 

kpaul:
i added a handler for dealing with unhandled exceptions (approach taken from here https://bytefish.de/blog/consistent_error_handling_with_nancy/). this way we can set a custom message and more importantly add the CORS headers to the response so that javascript applications can read the response. currently it only handles unhandled exceptions but in the future we can add handling for 404's and in the platform middleware we should be able to handle 403's. relevant code is in the class OnErrorHooks

i also moved the cors headers to the HttpResponseHeaders class so we can reuse them.

### DIFF
--- a/Fabric.Authorization.API/Infrastructure/PipelineHooks/OnErrorHooks.cs
+++ b/Fabric.Authorization.API/Infrastructure/PipelineHooks/OnErrorHooks.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using Fabric.Authorization.API.Constants;
+using Fabric.Authorization.API.Models;
+using Nancy;
+using Nancy.Responses.Negotiation;
+using Serilog;
+
+namespace Fabric.Authorization.API.Infrastructure.PipelineHooks
+{
+    public class OnErrorHooks
+    {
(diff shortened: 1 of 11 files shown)


Fabric.Authorization.API/Infrastructure/PipelineHooks/OnErrorHooks.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class OnErrorHooks
+    {
+        private readonly ILogger _logger;
+
+        public OnErrorHooks(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        internal dynamic HandleInternalServerError(NancyContext context, Exception exception,
```
michaelvidal24: Should we interrogate the environment variable that represents if this is a production vs debug environment, and if its a debug environment, return the stack trace in the error response?

kpaul: @michaelvidal24 i made the change to return the stack trace and the real exception message when running in development mode

PR #32, opened 2017-09-25
User/bsmithvt/group permission bug

bsmithvt:


Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -82,7 +84,9 @@ public class UsersModule : FabricModule<User>
             await SetDefaultRequest(userPermissionRequest);
             await CheckAccess(_clientService, userPermissionRequest.Grain, userPermissionRequest.SecurableItem,
                 AuthorizationReadClaim);
-            var groups = GetGroupsForAuthenticatedUser();
+
+            var subjectId = Context.CurrentUser.Claims.First(c => c.Type == Claims.Sub).Value;
```
michaelvidal24: Should we push this to a function in FabricModule?

bsmithvt: @michaelvidal24 , are you referring to the entire GetCurrentUserPermissions() method?

michaelvidal24: @bsmithvt, I was referring to just line 88:

```
var subjectId = Context.CurrentUser.Claims.First(c => c.Type == Claims.Sub).Value;
```

Seems like it is a candidate to be reused elswhere  in the authorization api so might be helpful to have in FabricModule.

PR #38, opened 2017-09-29
ensure subject is used instead of userid. 

kpaul:
get identity provider from claims when finding permissions for a user
when searching for user use the userid (subjectid:identityprovider)
added identityprovider to endpoint for getting a users groups

### DIFF
--- a/Fabric.Authorization.API/Modules/UsersMetadataModule.cs
+++ b/Fabric.Authorization.API/Modules/UsersMetadataModule.cs
@@ -22,7 +22,7 @@ namespace Fabric.Authorization.API.Modules
         private readonly Parameter _userIdParameter = new Parameter
         {
             Name = "userId",
-            Description = "UserId to use for the request",
+            Description = "SubjectId to use for the request",
             In = ParameterIn.Path,
             Required = true,
             Type = "integer"
--- a/Fabric.Authorization.API/Modules/UsersModule.cs
+++ b/Fabric.Authorization.API/Modules/UsersModule.cs
@@ -45,7 +45,7 @@ namespace Fabric.Authorization.API.Modules
                 async param => await this.AddGranularPermissions(param, denied: true).ConfigureAwait(false), null,
                 "AddDeniedPermissions");
 
-            Get("/{subjectId}/groups", async _ => await GetUserGroups().ConfigureAwait(false), null, "GetUserGroups");
+            Get("/{userId}/groups", async _ => await GetUserGroups().ConfigureAwait(false), null, "GetUserGroups");
         }
 
         private async Task<dynamic> GetUserGroups()
@@ -86,7 +86,7 @@ namespace Fabric.Authorization.API.Modules
                 AuthorizationReadClaim);
 
             var subjectId = SubjectId;
-            var groups = await GetGroupsForAuthenticatedUser(subjectId);
+            var groups = await GetGroupsForAuthenticatedUser(subjectId).ConfigureAwait(false);
 
             var permissions = await _permissionService.GetPermissionsForUser(
                 subjectId,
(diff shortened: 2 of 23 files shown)


Fabric.Authorization.API/Modules/UsersMetadataModule.cs
```diff
@@ -22,7 +22,7 @@ public class UsersMetadataModule : BaseMetadataModule
         private readonly Parameter _userIdParameter = new Parameter
         {
             Name = "userId",
```
michaelvidal24: Do we also need to change the name of the parameter from userId to subjectId?

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -37,31 +38,31 @@ public class UsersModule : FabricModule<User>
             Get("/permissions", async _ => await GetCurrentUserPermissions().ConfigureAwait(false), null,
                 "GetUserPermissions");
 
-            Post("/{userId}/AdditionalPermissions",
+            Post("/{subjectId}/AdditionalPermissions",
```
michaelvidal24: Do we also need identityProvider here? I think these endpoints could be used by actors who are not the subject/target of the permission changes.

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -37,31 +38,31 @@ public class UsersModule : FabricModule<User>
             Get("/permissions", async _ => await GetCurrentUserPermissions().ConfigureAwait(false), null,
                 "GetUserPermissions");
 
-            Post("/{userId}/AdditionalPermissions",
+            Post("/{subjectId}/AdditionalPermissions",
                 async param => await this.AddGranularPermissions(param, denied: false).ConfigureAwait(false), null,
                 "AddPermissions");
 
-            Post("/{userId}/DeniedPermissions",
+            Post("/{subjectId}/DeniedPermissions",
```
michaelvidal24: I think this endpoint has the same issue as above.

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -37,31 +38,31 @@ public class UsersModule : FabricModule<User>
...
             return groups;
         }
 
         private async Task<dynamic> AddGranularPermissions(dynamic param, bool denied)
         {
             var apiModel = this.Bind<GranularPermissionApiModel>();
 
-            if (apiModel.Target != param["userId"])
-                return CreateFailureResponse("Target must be the user id.", HttpStatusCode.BadRequest);
+            if (apiModel.Target != param["subjectId"])
```
michaelvidal24: Can we make this a case insensitive comparison? Also - not related to your changes, I am wondering why we make this check, i.e. what is the purpose of the Target property on the GranularPermissionApiModel? @deeiinu can you refresh my memory on the purpose of the Target property?

Fabric.Authorization.API/Modules/UsersModule.cs
kpaul: You are getting ahead of me here. That is what I am going to suggest in the api changes to support identity provider that I have not submitted to the team yet.

Fabric.Authorization.API/Modules/UsersMetadataModule.cs
kpaul: That is changed in another pull request where I have the swagger doc updates.

Fabric.Authorization.API/Modules/UsersModule.cs
kpaul: I was curious about that myself. I realize that it represents the subjectId of the user to change permissions for but I am not sure why its called Target.

PR #40, opened 2017-10-03
User/bsmithvt/func perf tests2

bsmithvt:
I had to bypass the HttpClientFactory in the Fabric.Platform package. I have TODOs to clean up the hard-coded logic that sets up the HttpClient. I'll probably do this by creating a new config section for the Fabric.Auth "client" that can be used as Mike suggested.

### DIFF
--- a/Fabric.Authorization.API/RemoteServices/Identity/Providers/IdentityServiceProvider.cs
+++ b/Fabric.Authorization.API/RemoteServices/Identity/Providers/IdentityServiceProvider.cs
@@ -1,13 +1,16 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Threading.Tasks;
 using Fabric.Authorization.API.Configuration;
 using Fabric.Authorization.API.Constants;
 using Fabric.Authorization.API.RemoteServices.Identity.Models;
 using Fabric.Platform.Http;
+using IdentityModel.Client;
 using Newtonsoft.Json;
 
 namespace Fabric.Authorization.API.RemoteServices.Identity.Providers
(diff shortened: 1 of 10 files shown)


Fabric.Authorization.API/RemoteServices/Identity/Providers/IdentityServiceProvider.cs
```diff
@@ -25,25 +28,49 @@ public IdentityServiceProvider(IHttpClientFactory httpClientFactory, IAppConfigu
...
                 ? new List<UserSearchResponse>()
                 : JsonConvert.DeserializeObject<IEnumerable<UserSearchResponse>>(await response.Content
                     .ReadAsStringAsync());
         }
+
+        private async Task<HttpClient> Create(string tokenUrl, string clientId, string clientSecret, Uri uri, string requestScope)
+        {
+            var tokenClient = new TokenClient(tokenUrl, clientId, clientSecret);
+            var response = await tokenClient.RequestClientCredentialsAsync(requestScope).ConfigureAwait(false);
+            return CreateWithAccessToken(uri, response.AccessToken);
```
michaelvidal24: Did the HttpClientFactory.CreateWithAccessToken method not work? Let me know, I can check it out and fix it up.

bsmithvt: @michaelvidal24 I can use it, but I still have to create the HttpClientFactory and TokenClient and manually. I've changed the code to use the factory since that does eliminate some repeated code,  but we still have some refactor work to do. The issue is the HttpClientFactory accepts a tokenUrl that is pulled from configuration at app startup, which at the moment is just "http://localhost:5001". We need to append "/connect/token" to that URL. As part of the refactor work to segregate config settings, we can probably fix all of this so that IdentityServiceProvider uses the injected HttpClientFactory. We are currently losing the correlationToken with how we're doing it, but that should fixed with the config refactor too.

bsmithvt: @michaelvidal24, the other issue is the clientId in appsettings.json is incorrect for the Fabric auth client.

PR #42, opened 2017-10-05
User/bsmithvt/user api bug fix

bsmithvt:
I have 1 more commit coming soon that will include additional logging in IdentitySearchService.

### DIFF
--- a/Fabric.Authorization.API/Modules/FabricModule.cs
+++ b/Fabric.Authorization.API/Modules/FabricModule.cs
@@ -71,7 +71,7 @@ namespace Fabric.Authorization.API.Modules
                 .WithHeader(HttpResponseHeaders.Location, selfLink);
         }
 
-        protected Negotiator CreateSuccessfulGetResponse<T1>(T1 model, HttpStatusCode statusCode = HttpStatusCode.Created)
+        protected Negotiator CreateSuccessfulGetResponse<T1>(T1 model, HttpStatusCode statusCode = HttpStatusCode.OK)
         {
             return Negotiate
                 .WithModel(model)
--- a/Fabric.Authorization.API/Modules/FabricModule.cs
+++ b/Fabric.Authorization.API/Modules/FabricModule.cs
@@ -71,6 +71,13 @@ namespace Fabric.Authorization.API.Modules
                 .WithHeader(HttpResponseHeaders.Location, selfLink);
         }
 
+        protected Negotiator CreateSuccessfulGetResponse<T1>(T1 model, HttpStatusCode statusCode = HttpStatusCode.Created)
+        {
+            return Negotiate
+                .WithModel(model)
+                .WithStatusCode(statusCode);
+        }
+
         protected Negotiator CreateFailureResponse(string message, HttpStatusCode statusCode)
         {
             var error = ErrorFactory.CreateError<T>(message, statusCode);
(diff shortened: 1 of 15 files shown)


Fabric.Authorization.API/Modules/IdentitySearchMetadataModule.cs
```diff
@@ -44,6 +44,11 @@ public IdentitySearchMetadataModule(ISwaggerModelCatalog modelCatalog, ISwaggerT
                         Code = (int) HttpStatusCode.OK,
                         Message = "OK"
                     },
+                    new HttpResponseMetadata<IEnumerable<IdentitySearchResponse>>
+                    {
+                        Code = (int) Nancy.HttpStatusCode.PartialContent,
+                        Message = "Partial success (e.g., results were found in Fabric.Authorization but the call out to Fabric.Identity failed)"
```
kpaul: nice addition. do you think callers will understand what data we get from Identity vs. Authorization?

Fabric.Authorization.API/Modules/FabricModule.cs
```diff
@@ -71,6 +71,13 @@ protected Predicate<Claim> AuthorizationManageClientsClaim
                 .WithHeader(HttpResponseHeaders.Location, selfLink);
         }
 
+        protected Negotiator CreateSuccessfulGetResponse<T1>(T1 model, HttpStatusCode statusCode = HttpStatusCode.Created)
```
michaelvidal24: For a GET should the default status Code be `Created` or `OK`?

Fabric.Authorization.API/Services/IdentitySearchService.cs
```diff
@@ -101,13 +119,13 @@ public class IdentitySearchService
...
                 await _identityServiceProvider.Search(request.ClientId, userList.Select(u => $"{u.SubjectId}:{u.IdentityProvider}"));
 
-            if (userDetails != null)
+            if (fabricIdentityUserResponse != null && fabricIdentityUserResponse.HttpStatusCode == System.Net.HttpStatusCode.OK)
             {
                 // update user details with Fabric.Identity response
-                foreach (var user in userDetails)
+                foreach (var user in fabricIdentityUserResponse.Results)
                 {
                     var userSearchResponse = userList.FirstOrDefault(u => u.SubjectId == user.SubjectId);
```
kpaul: should identity provider also be used here?

Fabric.Authorization.API/RemoteServices/Identity/Providers/IdentityServiceProvider.cs
```diff
@@ -52,13 +60,27 @@ public IdentityServiceProvider(IHttpClientFactory httpClientFactory, IAppConfigu
...
+            var response = await httpClient.PostAsync("api/users",
                 new StringContent(JsonConvert.SerializeObject(request), Encoding.UTF8, "application/json"));
 
-            return response.StatusCode != HttpStatusCode.OK
-                ? new List<UserSearchResponse>()
-                : JsonConvert.DeserializeObject<IEnumerable<UserSearchResponse>>(await response.Content
+            var results = new List<UserSearchResponse>();
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
+                _logger.Error($"Response from Fabric.Identity api/users => {response.StatusCode}");
```
michaelvidal24: Should we also log the content of the response, in case there is an error message from the remote service?

Fabric.Authorization.API/Modules/IdentitySearchMetadataModule.cs
bsmithvt: @kpaul , good point. Maybe I should explicitly call out which pieces of data come from Identity here so it's in the documentation.

Fabric.Authorization.API/Services/IdentitySearchService.cs
bsmithvt: @kpaul, not sure I follow?

kpaul: when searching for users returned from identity, will it be enough to only match on subject id or should we also match on identity provider

PR #43, opened 2017-10-05
update user endpoints to include identity provider in the url

kpaul:


Fabric.Authorization.Domain/Stores/Services/PermissionService.cs
```diff
@@ -139,6 +133,32 @@ public PermissionService(IPermissionStore permissionStore, RoleService roleServi
...
+
+                var deniedPermissionsToRemove = granularPermission.DeniedPermissions.Select(ap => ap.ToString()).ToList();
+                stored.DeniedPermissions = stored.DeniedPermissions.Where(p => !deniedPermissionsToRemove.Contains(p.ToString()));
+
+                await _permissionStore.AddOrUpdateGranularPermission(stored);
+
+            }
+            catch (NotFoundException<GranularPermission>)
+            {
+                // Do nothing
```
bsmithvt: Is there a reason we swallow this exception? Should we log the exception?

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -57,26 +58,52 @@ public class UsersModule : FabricModule<User>
...
+        {
+            var apiModel = this.Bind<GranularPermissionApiModel>();
+
+            foreach (var perm in apiModel.Permissions)
+                await CheckAccess(_clientService, perm.Grain, perm.SecurableItem, AuthorizationManageClientsClaim);
+
+            var granularPermissions = apiModel.ToGranularPermissionDomainModel();
+
+            granularPermissions.IdentityProvider = param.identityProvider;
+            granularPermissions.Id = param.subjectId;
```
bsmithvt: Why is the Id the subjectId and not a concatenation of the subjectId and identityProvider? If we want to get permissions for a user, subjectId and identityProvider will be required, right?

Fabric.Authorization.Domain/Models/GranularPermission.cs
```diff
@@ -23,11 +29,11 @@ public class GranularPermission : ITrackable, IIdentifiable, ISoftDelete
 
         public string CreatedBy { get; set; }
 
-        public string ModifiedBy { get; set; }
+        public string ModifiedBy { get; set; }        
 
         public override string ToString()
         {
-            return $"{Target}";
+            return $"{Identifier}";
```
bsmithvt: Should we include IdentityProvider here too?

Fabric.Authorization.Domain/Stores/Services/PermissionService.cs
```diff
@@ -118,17 +118,11 @@ public PermissionService(IPermissionStore permissionStore, RoleService roleServi
         {
             try
             {
-                var stored = await GetUserGranularPermissions(granularPermission.Target);
+                var stored = await GetUserGranularPermissions(granularPermission.Id);
```
bsmithvt: Should this include the IdentityProvider in the query (dependent on question above about Id being the SubjectId only)?

kpaul: yes i noticed that too and am going to change it

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -57,26 +57,52 @@ public class UsersModule : FabricModule<User>
...
 
-            if (denied)
-                granularPermissions.DeniedPermissions = apiModel.Permissions.Select(p => p.ToPermissionDomainModel());
-            else
-                granularPermissions.AdditionalPermissions =
-                    apiModel.Permissions.Select(p => p.ToPermissionDomainModel());
+            granularPermissions.IdentityProvider = param.identityProvider;
+            granularPermissions.Id = param.subjectId;
+
+            granularPermissions.DeniedPermissions = apiModel.Permissions
```
bsmithvt: If I'm not mistaken, clients will have to pass in 0 for "add" and 1 for "deny" since this is an enum. Should we make it a string so that it's easier for clients to remember? I noticed this with an enum I was using in a response and it was sending back integers for the enum.

Fabric.Authorization.Domain/Stores/Services/PermissionService.cs
kpaul: i did not think it seemed important enough to log. if the user has had no granular permissions added then there is nothing to delete and so not much to do.

michaelvidal24: I think it might be better to allow this exception to bubble up, and catch it in the module, and we can return a BadRequest with the appropriate error message so the consumer of the API knows that the user doesn't have any granular permissions defined.

Fabric.Authorization.API/Modules/UsersModule.cs
kpaul: Yeah thats a good point, it can be a concatenation of both

Fabric.Authorization.Domain/Models/GranularPermission.cs
kpaul: If Id becomes a concatenation of subject id and identity provider then it will be included here

Fabric.Authorization.API/Modules/UsersModule.cs
kpaul: Interesting. If thats the case then I will change it so the clients can pass in a string

Fabric.Authorization.Domain/Stores/Services/PermissionService.cs
kpaul: sure thing i can do that. BadRequest doesnt feel like the proper response code but I am having trouble thinking of a better one.

michaelvidal24: Yes, sorry, 404 would be better.

PR #44, opened 2017-10-06
1) added more logging, 2) do case-insensitive SubjectId comparison in…

bsmithvt:
… response from Identity


Fabric.Authorization.API/RemoteServices/Identity/Providers/IdentityServiceProvider.cs
```diff
@@ -66,15 +66,18 @@ public IdentityServiceProvider(IHttpClientFactory httpClientFactory, IAppConfigu
                 new StringContent(JsonConvert.SerializeObject(request), Encoding.UTF8, "application/json"));
 
             var results = new List<UserSearchResponse>();
+            var responseContent = await response.Content.ReadAsStringAsync();
```
michaelvidal24: Do we need to check for null here? I am not 100% sure if the Content property is null if no content was returned from the server, for example in the case where a 401 or 403 is returned.

PR #47, opened 2017-10-10
User/bsmithvt/permission bug

bsmithvt:


Fabric.Authorization.API/Modules/GroupsModule.cs
```diff
@@ -61,7 +61,7 @@ public class GroupsModule : FabricModule<Group>
             try
             {
                 this.RequiresClaims(AuthorizationReadClaim);
-                Group group = await _groupService.GetGroup(parameters.groupName);
+                Group group = await _groupService.GetGroup(parameters.groupName, ClientId);
```
michaelvidal24: We don't have to do this now, but at some point we will probably have to allow getting all the roles for a group regardless of the requesting clientid, most likely for the overall admin UI that will allow managing users and groups.

PR #51, opened 2017-10-11
update granular permission endpoints. remove granular permission api model

kpaul:
remove granular permission api model and update granular permission endpoints to accept a list of permission api models.

the ITrackable interface allows for the automatic setting of the created by/modified by user and date


Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -61,33 +61,47 @@ public class UsersModule : FabricModule<User>
 
         private async Task<dynamic> AddGranularPermissions(dynamic param)
         {
-            var apiModel = this.Bind<GranularPermissionApiModel>();            
+            var apiModel = this.Bind<List<PermissionApiModel>>();
```
michaelvidal24: After binding the model, should we check to see if the list is empty, and if so, let the consumer know that there were no permissions in the post? I've seen this happen when they try to pass something that is not an array of objects, and the nancy serializer just creates an empty list.

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -61,33 +61,47 @@ public class UsersModule : FabricModule<User>
...
 
-            await _permissionService.AddUserGranularPermissions(granularPermissions);
+            await _permissionService.AddUserGranularPermissions(granularPermission);
             return HttpStatusCode.NoContent;
         }      
 
         private async Task<dynamic> DeleteGranularPermissions(dynamic param)
         {
-            var apiModel = this.Bind<GranularPermissionApiModel>();
+            var apiModel = this.Bind<List<PermissionApiModel>>();
```
michaelvidal24: Same comment as above.

Fabric.Authorization.API/Modules/UsersModule.cs
kpaul: yeah, good call

PR #53, opened 2017-10-12
ensure allow and deny permission lists are mutually exclusive return error if not

kpaul:


Fabric.Authorization.Domain/Stores/Services/PermissionService.cs
```diff
@@ -148,21 +148,28 @@ public PermissionService(IPermissionStore permissionStore, RoleService roleServi
...
-            if (alreadyExistAllow.Any() || alreadyExistDeny.Any())
+            invalidPermissions.AddRange(allowPermissionsToAdd.Intersect(existingAllowPermissions)
+                .Select(p => new KeyValuePair<string,string>("Duplicate allow permissions", p.ToString())));
+            invalidPermissions.AddRange(denyPermissionsToAdd.Intersect(existingDenyPermissions)
+                .Select(p => new KeyValuePair<string, string>("Duplicate deny permissions", p.ToString() )));
+
+            invalidPermissions.AddRange(allowPermissionsToAdd.Intersect(existingDenyPermissions)
+                .Select(p => new KeyValuePair<string, string>("Exists as deny", p.ToString() )));
+            invalidPermissions.AddRange(denyPermissionsToAdd.Intersect(existingAllowPermissions)
+                .Select(p => new KeyValuePair<string, string>("Exists as allow", p.ToString())));
```
michaelvidal24: Do we also need to check if any of the new deny permissions overlap with the new allow permissions?

kpaul: do you mean if in a single request they add the same permission as both allow and deny?

PR #58, opened 2017-10-16
User/bsmithvt/get user permissions

bsmithvt:


bsmithvt:
Note this branch is misnamed. I initially thought it was going to 1 chunk of work for the granular permission ID case insensitivity work AND the new API to retrieve details about user's permissions. I think it should be 2 separate branches / PRs.

kpaul:
I dont think I can review until the conflicts are resolved

bsmithvt:
@kpaul, I've merged master back into this branch and resolved the conflicts, which were only in UserTests.cs.

PR #60, opened 2017-10-16
1) added better messaging around NotFoundExceptions and 2) gracefully…

bsmithvt:
… handle case in user search API where deleted group is mapped to a valid role

### DIFF
--- a/Fabric.Authorization.API/Services/IdentitySearchService.cs
+++ b/Fabric.Authorization.API/Services/IdentitySearchService.cs
@@ -63,7 +63,6 @@ namespace Fabric.Authorization.API.Services
 
             if (groupIds.Count == 0)
             {
-                _logger.Debug("groupIds is empty - returning empty result set.");
                 return new FabricAuthUserSearchResponse
                 {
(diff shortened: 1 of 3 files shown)


Fabric.Authorization.API/Services/IdentitySearchService.cs
```diff
@@ -74,9 +73,15 @@ public class IdentitySearchService
...
-                var group = await _groupService.GetGroup(groupId, request.ClientId);
-                groupEntities.Add(group);
+                try
+                {
+                    var group = await _groupService.GetGroup(groupId, request.ClientId);
+                    groupEntities.Add(group);
+                }
+                catch (NotFoundException<Group> ex)
+                {
+                    _logger.Error($"{ex.Message} (Group is mapped to at least 1 valid role)");
```
kpaul: is this error message correct? 

>  (Group is mapped to at least 1 valid role)

bsmithvt: @kpaul, yes, the query above gets all valid roles owned by the client and pulls the Groups out of those roles.

PR #66, opened 2017-10-24
User/bsmithvt/deleted group bug

bsmithvt:
I still need to add a couple of tests around Add->Delete->Add for groups with the same name.

### DIFF
--- a/Fabric.Authorization.Domain/Stores/CouchDB/CouchDBGroupStore.cs
+++ b/Fabric.Authorization.Domain/Stores/CouchDB/CouchDBGroupStore.cs
@@ -11,6 +11,8 @@ namespace Fabric.Authorization.Domain.Stores.CouchDB
 {
     public class CouchDbGroupStore : FormattableIdentifierStore<string, Group>, IGroupStore
     {
+        private const string IdDelimiter = "-:-:";
+
         public CouchDbGroupStore(
             IDocumentDbService dbService,
             ILogger logger,
@@ -21,7 +23,7 @@ namespace Fabric.Authorization.Domain.Stores.CouchDB
 
         public override async Task<Group> Add(Group group)
         {
-            // this will also catch older active records that do not have the unique identifier appended to the ID
+            // this will catch older active records that do not have the unique identifier appended to the ID
             var exists = await Exists(group.Id).ConfigureAwait(false);
             if (exists)
             {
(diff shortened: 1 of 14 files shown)


Fabric.Authorization.Domain/Stores/CouchDB/CouchDBGroupStore.cs
```diff
@@ -16,35 +19,82 @@ public class CouchDbGroupStore : FormattableIdentifierStore<string, Group>, IGro
...
+        {
+            // this will also catch older active records that do not have the unique identifier appended to the ID
+            var exists = await Exists(group.Id).ConfigureAwait(false);
+            if (exists)
+            {
+                throw new AlreadyExistsException<Group>($"Group ID {group.Id} already exists.");
+            }
+
+            // append unique identifier to document ID
+            group.Id = $"{group.Id}-{DateTime.UtcNow.Ticks}";
```
kpaul: a minor comment, we have been using : as the separator in document ids

bsmithvt:
@michaelvidal24,

Re: #1 and #2, those are good points. Since the ID is generated based on the `Name`, the checks I have in place should prevent this. For IDs with the new format, we do a query to see if any Group ID already exists that starts with the supplied `Name ` (including the delimiter to prevent Group1 and Group123 from being a false match), and there are already tests in place to ensure groups with the same name do not exist. The `Exists` method does a `Get` using this logic (first attempting an exact match failing over to a `starts with` match). Does it make sense to include this check in the Validator at the API layer since it will be done again at the `CouchDBStore` layer too? Below is the `Add` implementation in the CouchDBStore.

```
public override async Task<Group> Add(Group group)
        {
            // this will catch older active records that do not have the unique identifier appended to the ID
            var exists = await Exists(group.Id).ConfigureAwait(false);
            if (exists)
            {
                throw new AlreadyExistsException<Group>($"Group ID {group.Id} already exists.");
            }

            // append unique identifier to document ID
            group.Id = $"{group.Id}{IdDelimiter}{DateTime.UtcNow.Ticks}";
            return await Add(FormatId(group.Id), group);
        }
```

Re: #3, I believe the `AddGroup_AlreadyExists_BadRequest` test should cover this scenario.

michaelvidal24:
@bsmithvt - thanks for the explanation. I do think it makes sense to check this at the API layer as well, this way we can fail the request quickly and not waste resources processing something we know will fail down the line anyway.

bsmithvt:
@michaelvidal24, ok, I'll add that and a test to cover the scenario for old style IDs (since the tests will now generate groups with new style IDs). Also, I need to adjust the InMemoryGroupStore to behave the same as the CouchDBGroupStore (allow groups w/ the same name).

PR #70, opened 2017-10-26
User/bsmithvt/permission createdate bug

bsmithvt:


Fabric.Authorization.Domain/Stores/CouchDB/CouchDBPermissionStore.cs
```diff
@@ -66,13 +67,20 @@ public static CouchDbViews GetViews()
         {
             var userId = FormatId(granularPermission.Id);
             var perm = await DocumentDbService.GetDocument<GranularPermission>(userId);
+            var currentUser = GetActor();
 
             if (perm == null)
             {
+                granularPermission.Track(true, currentUser);
+                granularPermission.AdditionalPermissions = granularPermission.AdditionalPermissions.Track(true, currentUser);
```
michaelvidal24: I can see tracking the created date of the granularpermission document, but aren't the additional and denied permissions really references (or copies in this case) of permissions that live in their own document? If that is the case, shouldn't we be using the created/modified dates from those instances here?

bsmithvt: @michaelvidal24, I was viewing the create/modified dates in this context as the date the permission was added/modified to/in this user’s granular permission document and not the date the permission lookup was created for use by the system. Is this incorrect?

michaelvidal24: @bsmithvt - I am not sure what is 'correct' in this case. I can see a case for both ways. @lindachan465, what are your thoughts, should the Created/Updated date on the additional permission in the granular permission entity reflect the created/updated date that the source permission was created/updated, or should it reflect the date that the permission was added into the granular permissions allowed or denied collection?

lindachan465: I would say dates on a given model should reflect actions taken on that particular entity, so I would say the dates in granular permission should be when that granular permission was added/removed from the user.
Curious if we track these when we add a permission to a role or when we add groups to roles? If so, what do we do there? Should be consistent to avoid confusion.

michaelvidal24: @lindachan465  the role updated date time would get changed, but we would not change the created/updated date of the permission.

lindachan465: @michaelvidal24, had a quick chat with @bsmithvt and decided we will implement these to be consistent with how we are adding permissions -> roles and roles -> groups (which I believe is to use the source entity create/modified date instead of add/removal date). These are not critical fields that our implementer typically need for app logic, so until we decide to strip them out or find a real use case for them, keeping the implementation consistent for the time being would make most sense.

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -123,20 +123,64 @@ public class UsersModule : FabricModule<User>
...
                 await CheckAccess(_clientService, perm.Grain, perm.SecurableItem, AuthorizationManageClientsClaim);
+
+                // should never hit this with validation above, but the GetPermission call below
+                // complains with a possible InvalidOperationException
+                if (!perm.Id.HasValue)
+                {
+                    continue;
+                }
+
+                var permission = await _permissionService.GetPermission(perm.Id.Value);
```
michaelvidal24: Instead of querying for each permission in a loop, is it possible to query all the permissions in a single query so we don't have to potentially execute lots of queries here?

PR #72, opened 2017-10-26
ensure the correct status codes are returned for different scenarios

kpaul:
fixed a few places where we were returning the wrong status codes. 
added some tests to ensure the correct status codes are returned.

### DIFF
--- a/Fabric.Authorization.API/Modules/IdentitySearchModule.cs
+++ b/Fabric.Authorization.API/Modules/IdentitySearchModule.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.IO;
+using System.Text;
 using System.Threading.Tasks;
 using Fabric.Authorization.API.Configuration;
 using Fabric.Authorization.API.Models.Search;
@@ -7,7 +9,9 @@ using Fabric.Authorization.API.Services;
 using Fabric.Authorization.Domain.Exceptions;
 using Fabric.Authorization.Domain.Models;
 using Nancy;
+using Nancy.ErrorHandling;
 using Nancy.ModelBinding;
+using Nancy.Responses;
 using Nancy.Security;
 using Serilog;
 
@@ -50,11 +54,7 @@ namespace Fabric.Authorization.API.Modules
             catch (NotFoundException<Role> ex)
             {
                 return CreateFailureResponse(ex.Message, HttpStatusCode.NotFound);
-            }
-            catch (Exception ex)
-            {
-                return CreateFailureResponse(ex.Message, HttpStatusCode.InternalServerError);
-            }
+            }                 
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Fabric.Authorization.API/Modules/IdentitySearchModule.cs
```diff
@@ -50,11 +54,7 @@ public class IdentitySearchModule : SearchModule<IdentitySearchRequest>
             catch (NotFoundException<Role> ex)
             {
                 return CreateFailureResponse(ex.Message, HttpStatusCode.NotFound);
-            }
-            catch (Exception ex)
```
michaelvidal24: Was catching this general exception causing us to inadvertently swallow something?

kpaul: Yes, this was causing the model validation error to be swallowed so the client did not know the cause of the error.

Fabric.Authorization.IntegrationTests/Modules/UserTests.cs
```diff
@@ -1356,5 +1356,19 @@ public void TestUserWhitelist_Success()
             Assert.Contains("app/userprincipal.modifypatient", permissions.Permissions);
             Assert.Equal(3, permissions.Permissions.Count());
         }
+
+        [Fact]
+        [DisplayTestMethodName]
+        public void Test_GetPermissions_UserNotFound()
```
bsmithvt: Should this test be named "GetGroups" instead of "GetPermissions"?

Fabric.Authorization.API/Modules/IdentitySearchModule.cs
bsmithvt: I'm trying to remember why I was catching the general Exception here. Not sure if it had to do with not being able to reach Fabric.Identity or something going wrong with that call out or if it was related to the Content-Type being set on the GET request.

Fabric.Authorization.IntegrationTests/Modules/RolesTests.cs
```diff
@@ -233,5 +239,208 @@ public void TestDeleteRole_Fail(string id)
 
             Assert.Equal(HttpStatusCode.NotFound, delete.StatusCode);
         }
+
+        [Fact]
+        public void Test_DeletePermissionFromRole_PermissionDoesntExist_NotFoundException()
```
bsmithvt: Do we have a test that covers the scenario "delete a permission from role but the role doesn't contain the permission" (permission **does exist in the system, just not on the role)**?

kpaul: That is essentially what this is doing. We dont check the existence of the permission in the system on delete because we check that when its added to a role. So this check just ensures that the permission being deleted exists on the role

Fabric.Authorization.IntegrationTests/Modules/UserTests.cs
kpaul: Yes it should, good catch.

PR #77, opened 2017-10-30
update expected status codes for a couple functional tests

kpaul:


Fabric.Authorization.FunctionalTests/tests/authorizationtests.js
```diff
@@ -189,12 +189,12 @@ describe("authorization tests", function () {
     describe("register groups", function () {
         it("should return 400 for group HC Editor (already exists)", function () {
```
michaelvidal24: Should we update the message also?

PR #85, opened 2017-12-14
client and securable item entity models, mappers, and db context setup

kpaul:
i need to figure out the best way to model the securableItems collection on the securableItem entity. Its supposed to represent the children securable items. 

i also left a placeholder string for created by and modified by until we have a way to pull the authenticated users name. i will add a task for it.

### DIFF
--- a/Fabric.Authorization.Persistence.SqlServer/EntityModels/Client.cs
+++ b/Fabric.Authorization.Persistence.SqlServer/EntityModels/Client.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using Fabric.Authorization.Domain.Models;
+
+namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
+{
+    public class Client : ITrackable, ISoftDelete
+    {
+        public int Id { get; set; }
+        public string ClientId { get; set; }
+        public string Name { get; set; }
+        public int SecurableItemId { get; set; }
+        
+        public DateTime CreatedDateTimeUtc { get; set; }
+        public DateTime? ModifiedDateTimeUtc { get; set; }
+        public string CreatedBy { get; set; }
+        public string ModifiedBy { get; set; }
+        public bool IsDeleted { get; set; }
+
+        public SecurableItem TopLevelSecurableItem { get; set; }
+    }
+}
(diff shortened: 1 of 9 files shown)


Fabric.Authorization.Persistence.SqlServer/EntityModels/Client.cs
```diff
@@ -0,0 +1,21 @@
...
+using Fabric.Authorization.Domain.Models;
+
+namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
+{
+    public class Client : ITrackable, ISoftDelete
+    {
+        public int Id { get; set; }
+        public string ClientId { get; set; }
+        public string Name { get; set; }
+        public int SecurableItemId { get; set; }
```
bsmithvt: should this be nullable, or is it required when creating a client?

Fabric.Authorization.Persistence.SqlServer/EntityModels/Client.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Fabric.Authorization.Domain.Models;
+
+namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
+{
+    public class Client : ITrackable, ISoftDelete
+    {
+        public int Id { get; set; }
+        public string ClientId { get; set; }
+        public string Name { get; set; }
```
bsmithvt: should this be nullable, or is it required when creating a client?

kpaul: we validate it as a required field when posting a new one

kpaul: good catch, it should be nullable. i will update that

Fabric.Authorization.Persistence.SqlServer/EntityModels/SecurableItem.cs
```diff
@@ -0,0 +1,22 @@
...
+        public string Name { get; set; }
+
+        public DateTime CreatedDateTimeUtc { get; set; }
+        public DateTime? ModifiedDateTimeUtc { get; set; }
+        public string CreatedBy { get; set; }
+        public string ModifiedBy { get; set; }
+        public bool IsDeleted { get; set; }
+
+        public ICollection<SecurableItem> SecurableItems { get; set; }
+        public ICollection<Client> Clients { get; set; }
```
michaelvidal24: Shouldn't a SecurableItem only be related to a single Client?

kpaul: That is what I was not sure about. I can update the model to reflect the 1:1 relationship if it should be

PR #87, opened 2017-12-14
User/bsmithvt/ef group model

bsmithvt:
EF models for Group, Role, and GroupRole mapping. AutoMapper implementations and related User/Permission properties will come in a later PR.

### DIFF
--- a/Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
+++ b/Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
@@ -8,11 +8,11 @@ namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
     {
         public Group()
         {
-            Roles = new List<Role>();
+            GroupRoles = new List<GroupRole>();
             //Users = new List<User>();
         }
         public int Id { get; set; }
-        public Guid ExternalIdentifier { get; set; }
+        public string ExternalIdentifier { get; set; }
         public string Name { get; set; }
         public string Source { get; set; }
 
@@ -22,7 +22,7 @@ namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
         public string ModifiedBy { get; set; }
         public bool IsDeleted { get; set; }
 
-        public ICollection<Role> Roles { get; set; }
+        public ICollection<GroupRole> GroupRoles{ get; set; }
         //public ICollection<User> Users { get; set; }
     }
 }
--- a/Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
+++ b/Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using Fabric.Authorization.Domain.Models;
+
+namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
+{
+    public class Group : ITrackable, ISoftDelete
+    {
+        public Group()
+        {
+            Roles = new List<Role>();
+            //Users = new List<User>();
+        }
+        public int Id { get; set; }
+        public Guid ExternalIdentifier { get; set; }
+        public string Name { get; set; }
+        public string Source { get; set; }
+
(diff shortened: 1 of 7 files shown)


Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public class Group : ITrackable, ISoftDelete
+    {
+        public Group()
+        {
+            Roles = new List<Role>();
+            //Users = new List<User>();
+        }
+        public int Id { get; set; }
+        public Guid ExternalIdentifier { get; set; }
```
michaelvidal24: I think we set this value to GroupName in the API, so I don't think it can be a Guid.

Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
```diff
@@ -0,0 +1,28 @@
...
+        public string Source { get; set; }
+
+        public DateTime CreatedDateTimeUtc { get; set; }
+        public DateTime? ModifiedDateTimeUtc { get; set; }
+        public string CreatedBy { get; set; }
+        public string ModifiedBy { get; set; }
+        public bool IsDeleted { get; set; }
+
+        public ICollection<Role> Roles { get; set; }
+        //public ICollection<User> Users { get; set; }
```
michaelvidal24: Are we leaving this commented out because we haven't gotten to Users yet?

Fabric.Authorization.Persistence.SqlServer/EntityModels/GroupRole.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using Fabric.Authorization.Domain.Models;
+
+namespace Fabric.Authorization.Persistence.SqlServer.EntityModels
+{
+    public class GroupRole : ITrackable, ISoftDelete
```
michaelvidal24: Is this an explicit join entity? Does EF core not support many to many relationships with modelling the explicit join entity? Maybe a better question is, what are the advantages vs disadvantages of an explicit join entity?

Fabric.Authorization.Persistence.SqlServer/EntityModels/Role.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public Role()
+        {
+            Groups = new List<Group>();
+            ChildRoles = new List<Role>();
+            /*Permissions = new List<Permission>();
+            DeniedPermissions = new List<Permission>();*/
+        }
+        public int Id { get; set; }
+        public int? ParentRoleId { get; set; }
```
michaelvidal24: The domain model requires a Guid for ParentRole, I guess we'll have to handle that in the model mapping.

Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
bsmithvt: Yes

Fabric.Authorization.Persistence.SqlServer/EntityModels/GroupRole.cs
bsmithvt: Just to close the loop in this thread, below are 2 articles referencing EF Core many to many relationships. I do need to adjust my models to reference the join/linking entity as well. I will do that.

https://docs.microsoft.com/en-us/ef/core/modeling/relationships
Many-to-many relationships without an entity class to represent the join table are not yet supported. However, you can represent a many-to-many relationship by including an entity class for the join table and mapping two separate one-to-many relationships.

http://www.thereformedprogrammer.net/updating-many-to-many-relationships-in-entity-framework-core/

Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
bsmithvt: I've addressed this.

bsmithvt:
@kpaul, yes, I will do that next and the automapper stuff.

PR #92, opened 2017-12-15
User/bsmithvt/group role modelbuilder automapper

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Fabric.Authorization.Persistence.SqlServer.csproj
```diff
@@ -21,7 +21,6 @@
   <ItemGroup>
     <Folder Include="DependencyInjection\" />
     <Folder Include="Configuration\" />
-    <Folder Include="Mappers\" />
```
kpaul: did the Mappers folder get excluded?

Fabric.Authorization.Persistence.SqlServer/Mappers/UserMapperProfile.cs
```diff
@@ -0,0 +1,8 @@
+﻿using AutoMapper;
+
+namespace Fabric.Authorization.Persistence.SqlServer.Mappers
+{
+    public class UserMapperProfile : Profile
+    {
+    }
```
kpaul: was this intentionally left blank?

Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
```diff
@@ -9,11 +9,11 @@ public class Group : ITrackable, ISoftDelete
         public Group()
         {
             GroupRoles = new List<GroupRole>();
-            Users = new List<User>();
+            UserGroups = new List<UserGroup>();
```
michaelvidal24: Should we be consistent with the naming convention when referencing join entities?

bsmithvt: Where is the inconsistency?

Fabric.Authorization.Persistence.SqlServer/Mappers/UserMapperProfile.cs
bsmithvt: Yes, I will remove the file.

Fabric.Authorization.Persistence.SqlServer/Fabric.Authorization.Persistence.SqlServer.csproj
bsmithvt: Not sure why that keeps happening locally for me. I’ll fix.

kpaul: check the project file and make sure there is not an ItemGroup section that is removing it. I saw it in there before and removed it

Fabric.Authorization.Persistence.SqlServer/EntityModels/Group.cs
michaelvidal24: We use GroupRoles in one case, and UserGroups in another case. It might be easier if we are consistent with the order if possible, i.e. parent then children, e.g. GroupRoles and GroupUsers...

Fabric.Authorization.Persistence.SqlServer/Mappers/UserMapperProfile.cs
bsmithvt: Actually, I was wrong here. I needed to add a constructor that configured the mapping. That is done and checked in.

PR #96, opened 2017-12-18
add references to api project and sql server project to run migrations with api as startup project

kpaul:


bsmithvt:
@kpaul, I think you can remove the `netcoreapp1.1;` from the project file, right?

```
<PropertyGroup>
    <TargetFrameworks>netcoreapp1.1;netstandard1.3</TargetFrameworks>
  </PropertyGroup>
```

PR #100, opened 2017-12-19
User/bsmithvt/sql server permission store

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Fabric.Authorization.Persistence.SqlServer.csproj
```diff
@@ -33,7 +33,6 @@
     <Folder Include="DependencyInjection\" />
     <Folder Include="Configuration\" />
     <Folder Include="Mappers\" />
-    <Folder Include="Stores\" />
```
kpaul: I wonder why this keeps happening. It used to be the Mappers folder and now its the Stores folder.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            _authorizationDbContext.Permissions.Add(entity);
+            await _authorizationDbContext.SaveChangesAsync();
+            return model;
+        }
+
+        public async Task<Permission> Get(Guid id)
+        {
+            var permission = await _authorizationDbContext.Permissions
+                .SingleOrDefaultAsync(p =>
```
kpaul: Do any of the navigation properties need to be included in the query?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -0,0 +1,129 @@
...
+                    p.PermissionId == model.Id
+                    && !p.IsDeleted);
+
+            if (permission == null)
+            {
+                throw new NotFoundException<User>($"Could not find {typeof(Permission).Name} entity with ID {model.Id}");
+            }
+
+            permission.IsDeleted = true;
```
kpaul: should the RolePermission and UserPermission entities related to this permission also be marked as deleted?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            var idParts = SplitUserId(granularPermission.Id);
+
+            var user = await _authorizationDbContext.Users
+                .Include(u => u.UserPermissions)
+                .ThenInclude(up => up.Permission)
+                .SingleOrDefaultAsync(u => u.IdentityProvider.Equals(idParts[0], StringComparison.OrdinalIgnoreCase)
+                                           && u.SubjectId.Equals(idParts.Length > 1 ? idParts[1] : idParts[0], StringComparison.OrdinalIgnoreCase)
+                                           && !u.IsDeleted);
+
```
kpaul: i think there is a null check missing for user

Fabric.Authorization.Persistence.SqlServer/Fabric.Authorization.Persistence.SqlServer.csproj
bsmithvt: Yeah it's weird. I'll fix this.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
bsmithvt: In this context I don't think so. I think when retrieving groups/roles and users we will pull in the corresponding permissions, but I envisioned this as just a lookup resource to display permissions. Perhaps @michaelvidal24 can comment on this.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
bsmithvt: That's a good question. Perhaps the queries that retrieve the RolePermission and UserPermission entities should ensure the corresponding Permission is not deleted?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
bsmithvt: Good catch - I'll fix this.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
kpaul: Which queries are you referring to? The Get method in this class?

bsmithvt: I meant the queries that are in the SqlServerUserStore and SqlServerRoleStore.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
michaelvidal24: I think that makes sense, unless/until we need to pull in all the nav properties we shouldn't.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
kpaul: Ahh ok. I think all Get methods should check that the entity being searched for is not marked deleted. My original question here though relates more to the fact that the UserPermission and RolePermission entities implement ITrackable so shouldnt we be marking them deleted as well?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                permissions = permissions.Where(p => string.Equals(p.SecurableItem.Name, securableItem));
+            }
+
+            if (!string.IsNullOrWhiteSpace(permissionName))
+            {
+                permissions = permissions.Where(p => string.Equals(p.Name, securableItem));
+            }
+
+            return Task.FromResult(permissions.Select(p => p.ToModel()).AsEnumerable());
```
michaelvidal24: Do we need to check the IsDeleted flag here?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
bsmithvt: Yeah, you're probably right now that I think about it. I just meant for the User#Get, we only pull in related Permission entities that aren't deleted, but it's probably best to just do like you're doing in the group store and mark all relationship entities as deleted as well.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -0,0 +1,210 @@
...
+                .Where(p => string.Equals(p.Grain, grain, StringComparison.OrdinalIgnoreCase));
+
+            if (!string.IsNullOrWhiteSpace(securableItem))
+            {
+                permissions = permissions.Where(p => string.Equals(p.SecurableItem.Name, securableItem));
+            }
+
+            if (!string.IsNullOrWhiteSpace(permissionName))
+            {
+                permissions = permissions.Where(p => string.Equals(p.Name, securableItem));
```
michaelvidal24: Should we be comparing against the permissionName here and not the securableItem?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -0,0 +1,210 @@
...
+                                           && u.SubjectId.Equals(idParts.Length > 1 ? idParts[1] : idParts[0], StringComparison.OrdinalIgnoreCase)
+                                           && !u.IsDeleted);
+
+            if (user == null)
+            {
+                throw new NotFoundException<User>($"Could not find {typeof(User).Name} User ID {granularPermission.Id}");
+            }
+
+            // remove all current permissions first and then replace them with the new set of permissions
+            var currentUserPermissions = user.UserPermissions.Where(up => !up.IsDeleted);
```
michaelvidal24: Is there a better way to do this than removing and readding all the permissions? I am concerned that EF will generate a bunch of deletes/inserts where only a few or one is required.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
bsmithvt: Thanks - copy/paste error.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
bsmithvt: Yes, fixed.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
michaelvidal24: We don't have to solve this with this PR, but it might be something we want to look at in the future.

bsmithvt: Ok, I was starting to think about it...I was mimicking the CouchDB behavior for now, but I agree there is probably a better way to do and don't love the current approach.

PR #101, opened 2017-12-19
added sql server group store

kpaul:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
```diff
@@ -0,0 +1,118 @@
...
+            model.ToEntity(group);
+
+            _authorizationDbContext.Groups.Update(group);
+            await _authorizationDbContext.SaveChangesAsync();
+        }
+
+        public async Task<bool> Exists(string id)
+        {
+            var group = await _authorizationDbContext.Groups
+                .SingleOrDefaultAsync(g => g.GroupId.Equals(id, StringComparison.OrdinalIgnoreCase)
```
bsmithvt: Should this be checking the Name property also?

kpaul: Good call. Looking back over the group model I think name is the correct key to use when querying.

PR #102, opened 2017-12-19
added SqlServerRoleStore

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                rolePermission.IsDeleted = true;
+            }
+
+            foreach (var groupRole in role.GroupRoles)
+            {
+                groupRole.IsDeleted = true;
+            }
+
+            await _authorizationDbContext.SaveChangesAsync();
```
michaelvidal24: Should we consider wrapping this in a try/catch so we can deal with concurrency exceptions? In the case of a concurrency exception, we may want to retry the update. Not sure if we want it at this level, or if we want to handle it in the dbcontext, but something to think about.

bsmithvt: True - the same would apply to the permission store where we loop to delete the related entities. I'll hold off on adding it here until we decide whether we want to add it in the dbcontext.

PR #104, opened 2017-12-20
moved db specific stuff out of the GroupService class and into the store implementations

kpaul:
one thing about this approach is i had to move all the code out of the service and into the store methods. the service class basically becomes a pass through. this is because the store methods return domain models and i did not want to have to query again for entities in the sql server store implementation. we can discuss tomorrow morning


bsmithvt:
I'm not sure of everyone else's thoughts, but even though the service is currently just a pass through (in this case), I'm hesitant to suggest that we eliminate it altogether because there may be scenarios in the future where we do need logic to go in there, and since we already have that layer and it's wired up in the modules, my vote would be to leave it for now.

PR #108, opened 2017-12-20
User/bsmithvt/refactor role service

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -30,8 +34,18 @@ public SqlServerRoleStore(IAuthorizationDbContext authorizationDbContext)
...
+        {
+            var role = await GetEntityModel(id);
+            return role.ToModel();
+        }
+
+        public async Task<EntityModels.Role> GetEntityModel(Guid id)
         {
             var role = await _authorizationDbContext.Roles
+                .Include(r => r.RolePermissions)
+                .ThenInclude(rp => rp.Permission)
```
michaelvidal24: Do we need to filter to eliminate deleted permissions and permissions that are deleted from the role? Same comment with groups.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -34,6 +34,12 @@ public SqlServerRoleStore(IAuthorizationDbContext authorizationDbContext, IPermi
         }
 
         public async Task<Role> Get(Guid id)
+        {
+            var role = await GetEntityModel(id);
+            return role.ToModel();
+        }
+
+        public async Task<EntityModels.Role> GetEntityModel(Guid id)
```
kpaul: should this method be public?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -41,16 +55,20 @@ public SqlServerRoleStore(IAuthorizationDbContext authorizationDbContext)
...
             }
 
-            return role.ToModel();
+            return role;
         }
 
         public async Task<IEnumerable<Role>> GetAll()
         {
             var roles = await _authorizationDbContext.Roles
+                .Include(r => r.RolePermissions)
```
michaelvidal24: Same comments as above.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+            }
+
+            return roles;
+        }
+
+        public async Task<Role> AddPermissionsToRole(Role role, ICollection<Permission> permissions)
+        {
+            var roleEntity = await _authorizationDbContext.Roles
+                .SingleOrDefaultAsync(r =>
+                    r.RoleId == role.Id
```
michaelvidal24: Do we need to include the list of permissions here so we can verify that the permission hasn't already been added to the role? Or is the check being done in the service?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+                    && !r.IsDeleted);
+
+            if (role == null)
+            {
+                throw new NotFoundException<Role>($"Could not find {typeof(Role).Name} entity with ID {role.Id}");
+            }
+
+            var permissionIds = permissions.Select(p => p.Id);
+            var permissionDictionary = _authorizationDbContext.Permissions
```
michaelvidal24: Do we need to run this query here, we already have the PermissionId's from the service?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+        }
+
+        public async Task<Role> AddPermissionsToRole(Role role, ICollection<Permission> permissions)
+        {
+            var roleEntity = await _authorizationDbContext.Roles
+                .SingleOrDefaultAsync(r =>
+                    r.RoleId == role.Id
+                    && !r.IsDeleted);
+
+            if (role == null)
```
michaelvidal24: Should this be role, or roleEntity?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+                {
+                    RoleId = roleEntity.Id,
+                    PermissionId = permissionDictionary[permission.Id].Id,
+                    PermissionAction = PermissionAction.Allow
+                });
+            }
+
+            await _authorizationDbContext.SaveChangesAsync();
+
+            var updatedRole = await Get(role.Id);
```
michaelvidal24: Do we need to get the role again? We already have a reference to it don't we?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+
+        public async Task<Role> RemovePermissionsFromRole(Role role, Guid[] permissionIds)
+        {
+            var roleEntity = await _authorizationDbContext.Roles
+                .Include(r => r.RolePermissions)
+                .SingleOrDefaultAsync(r =>
+                    r.RoleId == role.Id
+                    && !r.IsDeleted);
+
+            if (role == null)
```
michaelvidal24: Should this be role or roleEntity?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+            {
+                throw new NotFoundException<Role>($"Could not find {typeof(Role).Name} entity with ID {role.Id}");
             }
 
-            return Task.FromResult(roles.Select(p => p.ToModel()).AsEnumerable());
+            var permissionDictionary = _authorizationDbContext.Permissions
+                .Where(p => !p.IsDeleted)
```
michaelvidal24: Similar comment as above, do we need to query for these id's?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -118,15 +142,101 @@ public Task<IEnumerable<Role>> GetRoles(string grain, string securableItem = nul
...
+                var rolePermissionToRemove = roleEntity.RolePermissions.Single(
+                    rp => rp.RoleId == roleEntity.Id
+                          && rp.PermissionId == permissionDictionary[permissionId].Id);
+
+                roleEntity.RolePermissions.Remove(rolePermissionToRemove);
+            }
+
+            await _authorizationDbContext.SaveChangesAsync();
+
+            var updatedRole = await Get(role.Id);
```
michaelvidal24: Not sure if we need to query to get the role, we already have it.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: We do need filter out deleted RolePermission entities now that we're doing a soft delete.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: The service is doing the check (I left it there so it would be done for all stores).

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: roleEntity

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: We have the PermissionIds, which is a set of Guids. I need to get the Sql Server `int` ID to create the relationship.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: I retrieve it again to ensure the Domain model is refreshed correctly before returning. It has collections of Allowed and Denied permissions. I suppose I could pull these entities out of the existing dictionary, transform them to the Permission domain model, and add them to the appropriate collection.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: roleEntity - fixed

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: Same response - I need the Sql Server `int` ID.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: This has been fixed.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
michaelvidal24: Wouldn't the role that was passed just have all the references it needs? I am concerned about the number of queries we are doing in general.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: No, this returns the entity model. I'm using it internally only so that I don't have this same code repeated in other places internally.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
kpaul: If the parent is not included in the result then I dont think the related entities will get loaded

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
michaelvidal24: OK - If having to requery to make an association is a side effect, I think we need to consider the idea of getting rid of the Sql Server identity column as an int. We don't have to consider it for this PR, but perhaps in another PR.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: The role that is passed in does not yet have the new permissions added. Perhaps I can just add the permission to the passed-in Role's permissions collection. I'll do that to eliminate that query.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
bsmithvt: Not sure what you mean?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
michaelvidal24: Yeah - lets add a work item to the PBI to investigate and potentially make that change across the entities.

PR #114, opened 2018-01-04
User/bsmithvt/sql server bootstrapper

bsmithvt:
I wanted to get this merged so that Sql Server integration tests can be started. The next steps are:

1) Move the CouchDB and InMemory code to their respective projects.
2) Push all of the DB-specific registration code out of the API project and into the DB-specific projects.

### DIFF
--- a/Fabric.Authorization.API/Bootstrapper.cs
+++ b/Fabric.Authorization.API/Bootstrapper.cs
@@ -12,6 +12,7 @@ using Fabric.Authorization.Domain.Events;
 using Fabric.Authorization.Domain.Stores;
 using Fabric.Authorization.Domain.Stores.CouchDB;
 using Fabric.Authorization.Domain.Stores.InMemory;
+using Fabric.Authorization.Persistence.SqlServer.Services;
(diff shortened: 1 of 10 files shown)


Fabric.Authorization.API/Bootstrapper.cs
```diff
@@ -144,7 +145,7 @@ private void ConfigureSingletonRegistrations(TinyIoCContainer container)
             ConfigureSingletonDataStores(container, _appConfig);
         }
 
-        private void ConfigureSingletonDataStores(TinyIoCContainer container, IAppConfiguration appConfig)
+        private static void ConfigureSingletonDataStores(TinyIoCContainer container, IAppConfiguration appConfig)
```
kpaul: what was the reason for making this static?

bsmithvt: @kpaul I was just seeing a Resharper warning / suggestion to make it static.

kpaul: ahh ok. maybe it should be extracted out into an extension method

PR #115, opened 2018-01-04
move dbcontext connection string to app config 

kpaul:
I am not sure where to register the options so we can set the connection string via the DI container.


bsmithvt:
@kpaul, are you referring to registering `IConnectionStrings`? Would it be in `ConfigureSingletonRegistrations`?

PR #121, opened 2018-01-09
User/mvidal/bootstrap db tests

michaelvidal24:


michaelvidal24:
Note that the database creation code is currently commented out since we don't have a sql server setup in the build pipeline yet, and the tests will fail without one. Once we have the Sql Server in the build pipeline we can uncomment the call to the database creation code.

PR #125, opened 2018-01-10
Updates to fix user tests that don't depend on groups.

michaelvidal24:
This PR will fix user tests that don't rely on groups or roles. Once the groups tests are fixed up I suspect the other tests will succeed, but we'll have to verify.


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerPermissionStore.cs
```diff
@@ -204,15 +212,15 @@ public SqlServerPermissionStore(IAuthorizationDbContext authorizationDbContext)
...
-        /// <returns></returns>
-        private static string[] SplitUserId(string id)
+        private static string[] SplitGranularPermissionId(string id)
         {
-            var delimiter = new[] { @"\" };
-            return id.Split(delimiter, StringSplitOptions.RemoveEmptyEntries);
+            var delimiter = new[] { @":" };
+            var idParts = id.Split(delimiter, StringSplitOptions.RemoveEmptyEntries);
+            if (idParts.Length != 2)
+            {
```
bsmithvt: Shouldn't the message be `{identityProvider}:{subjectId}`?

michaelvidal24: No - for the granular permission the id is subjectid:identityprovider - that was one of the bugs i had to fix.

michaelvidal24: Actually, good catch - I had it incorrect, it should have been {subjectId}:{identityProvider}, I've made the change.

PR #130, opened 2018-01-10
User/bsmithvt/sqlserver userstore fixes

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerUserStore.cs
```diff
@@ -113,7 +113,7 @@ public SqlServerUserStore(IAuthorizationDbContext authorizationDbContext)
 
         private static string[] SplitId(string id)
         {
-            var delimiter = new [] {@"\"};
+            var delimiter = new [] {@":"};
```
michaelvidal24: Should we throw an ArgumentException here if there aren't two parts to the id?

bsmithvt: Yes, good idea.

PR #131, opened 2018-01-10
Fix up setting relationship to securable item when saving a permissio…

michaelvidal24:
…n and role. Fix automapper mappings. Include additional tables when querying for roles.


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -138,6 +139,10 @@ private IEnumerable<EntityModels.Role> GetRoleEntityModels(string grain, string
         {
             var roles = _authorizationDbContext.Roles
                 .Include(r => r.SecurableItem)
```
bsmithvt: Do we need to exclude deleted related entities here?

michaelvidal24: Yeah - i'll fix that in a later PR.

PR #147, opened 2018-02-12
User/mvidal/refactor check access

michaelvidal24:
Added the ability to create an admin role with the authorization admin permission when a new client is created.
Added a ClientOwner column to SecurableItem so that it is easier to check if the current client owns a securable item in the case of a shared grain/securable item.
Added a config setting that will allow specifying a default user/group to assign to the automatically created admin role.
Refactored the CheckAccess function into separate CheckReadAccess and CheckWriteAccess functions to reduce the logic the single function was performing.


Fabric.Authorization.API/Modules/FabricModule.cs
```diff
@@ -94,27 +95,34 @@ protected Negotiator CreateFailureResponse(IEnumerable<string> messages, HttpSta
...
         {
             string grainAsString = grain.ToString();
             string securableItemAsString = securableItem.ToString();
-            var grainModel = await clientService.GetGrain(grainAsString);
-            if (grainModel != null && grainModel.IsShared)
+
+            if (HasSubjectId)
             {
-                await AccessService.CheckSharedAccess(grainModel, securableItemAsString, isWriteOperaion, this);
+                var grainModel = await clientService.GetGrain(grainAsString);
```
bsmithvt: What if grainModel is null?

PR #151, opened 2018-02-14
Added a POST /user endpoint

michaelvidal24:
So we can add a user before we assign roles to the user.
Created corresponding tests and swagger docs.

### DIFF
--- a/Fabric.Authorization.UnitTests/Users/UsersModuleTests.cs
+++ b/Fabric.Authorization.UnitTests/Users/UsersModuleTests.cs
@@ -28,7 +28,7 @@ namespace Fabric.Authorization.UnitTests.Users
                 .SetupGroupExists(_existingGroups);
 
             _mockUserStore = new Mock<IUserStore>()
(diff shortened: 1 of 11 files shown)


Fabric.Authorization.UnitTests/Users/UsersModuleTests.cs
```diff
@@ -99,6 +99,106 @@ public void GetUserPermissions_NoParameters_DefaultsToTopLevel()
...
+
+            var postResponse = usersModule.Post("/user", with =>
+            {
+                with.HttpRequest();
+                with.JsonBody(new
+                {
+                    SubjectId = subjectId,
+                    IdentityProvider = identityProvider
+                });
+            }).Result;
```
kpaul: is it ok to use .Result here?

michaelvidal24: The unit test use Mock stores which aren't async, so we aren't getting deadlocks here. We should convert them to async, just haven't yet.

PR #152, opened 2018-02-16
Added POST /user/{identityProvider}/{subjectId}/roles

michaelvidal24:
Also added a basic GET /user/{identityProvider}/{subjectId}


Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -68,6 +79,28 @@ public class UsersModule : FabricModule<User>
...
+        private async Task<dynamic> GetUser(dynamic param)
+        {
+            CheckReadAccess();
+            try
+            {
+                User user = await _userService.GetUser(param.subjectId, param.identityProvider);
+                return user.ToUserApiModel();
+            }
+            catch (NotFoundException<User>)
+            {
```
bsmithvt: Was an empty string for the message intentional?

Fabric.Authorization.Domain/Services/UserService.cs
```diff
@@ -8,10 +11,17 @@ namespace Fabric.Authorization.Domain.Services
     public class UserService
     {
         private readonly IUserStore _userStore;
+        private readonly IRoleStore _roleStore;
 
-        public UserService(IUserStore userStore)
+        public UserService(IUserStore userStore, IRoleStore roleStore)
```
bsmithvt: Do we need not null validations on both constructor parameters?

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -68,6 +79,28 @@ public class UsersModule : FabricModule<User>
...
+            Post("/{identityProvider}/{subjectId}/roles",
+                async param => await AddRolesToUser(param).ConfigureAwait(false), null,
+                "AddRolesToUser");
+        }
+
+        private async Task<dynamic> GetUser(dynamic param)
+        {
+            CheckReadAccess();
+            try
+            {
```
bsmithvt: Verify `subjectId` and `identityProvider` are not null?

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -270,6 +303,58 @@ public class UsersModule : FabricModule<User>
                     HttpStatusCode.NotFound);
             }
         }
+
+        private async Task<dynamic> GetUserRoles()
+        {
+            this.RequiresClaims(AuthorizationReadClaim);
+            var roleUserRequest = this.Bind<RoleUserRequest>();
+            try
+            {
```
bsmithvt: Verify `subjectId` and `identityProvider` are not null?

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -270,6 +303,58 @@ public class UsersModule : FabricModule<User>
...
+
+        private async Task<dynamic> AddRolesToUser(dynamic param)
+        {
+            var apiRoles = this.Bind<List<RoleApiModel>>();
+            var errorResponse = await ValidateRolesToAdd(apiRoles);
+            if (errorResponse != null)
+            {
+                return errorResponse;
+            }
+
```
bsmithvt: Verify `subjectId` and `identityProvider` are not null?

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -284,5 +369,24 @@ public class UsersModule : FabricModule<User>
...
+        {
+            if (apiRoles.Count == 0)
+            {
+                return await CreateFailureResponse(
+                    InvalidRoleArrayMessage, HttpStatusCode.BadRequest);
+            }
+
+            if (apiRoles.Any(r => !r.Id.HasValue || r.Id.Value == Guid.Empty || string.IsNullOrEmpty(r.Grain) ||
+                                  string.IsNullOrEmpty(r.SecurableItem)))
+            {
```
bsmithvt: Can we put this line before the `if` and then use the `messages` variable in the `if` so we're not executing this query twice?

Fabric.Authorization.Domain/Services/UserService.cs
```diff
@@ -20,6 +30,12 @@ public UserService(IUserStore userStore)
             return user != null ? user.Groups : new List<string>();
         }
 
+        public async Task<ICollection<Role>> GetRolesForUser(string subjectId, string identityProvider)
+        {
```
bsmithvt: Handle a `NotFoundException<User>` here?

Fabric.Authorization.Domain/Services/UserService.cs
```diff
@@ -29,5 +45,38 @@ public UserService(IUserStore userStore)
         {
             return await _userStore.Exists($"{subjectId}:{identityProvider}");
         }
+
+        public async Task<User> AddRolesToUser(IList<Role> rolesToAdd, string subjectId, string identityProvider)
+        {
```
bsmithvt: Handle a `NotFoundException<User>` here?

Fabric.Authorization.Domain/Services/UserService.cs
```diff
@@ -29,5 +45,38 @@ public UserService(IUserStore userStore)
...
+            var user = await _userStore.Get($"{subjectId}:{identityProvider}");
+            var grainSecurableItems = rolesToAdd.Select(r => new Tuple<string, string>(r.Grain, r.SecurableItem))
+                .Distinct();
+            var existingRoles = new List<Role>();
+            foreach (var tuple in grainSecurableItems)
+            {
```
bsmithvt: Handle `NotFoundException<Role>` here?

Fabric.Authorization.Domain/Services/UserService.cs
```diff
@@ -8,10 +11,17 @@ namespace Fabric.Authorization.Domain.Services
...
 
-        public UserService(IUserStore userStore)
+        public UserService(IUserStore userStore, IRoleStore roleStore)
         {
             _userStore = userStore;
+            _roleStore = roleStore;
+        }
+
+        public async Task<User> GetUser(string subjectId, string identityProvider)
+        {
```
bsmithvt: Handle a `NotFoundException<User>` here?

Fabric.Authorization.Domain/Services/UserService.cs
```diff
@@ -29,5 +45,38 @@ public UserService(IUserStore userStore)
...
+                .Distinct();
+            var existingRoles = new List<Role>();
+            foreach (var tuple in grainSecurableItems)
+            {
+                existingRoles.AddRange(await _roleStore.GetRoles(tuple.Item1, tuple.Item2));
+            }
+
+            var exceptions = new List<Exception>();
+            foreach (var role in rolesToAdd)
+            {
```
bsmithvt: Should we use the `role.Name` in the error messages below?

Fabric.Authorization.API/Modules/UsersModule.cs
michaelvidal24: I don't think they can be null, since it wouldn't be the route without those params, and the NotFoundException should catch this one anyway.

Fabric.Authorization.API/Modules/UsersModule.cs
michaelvidal24: I'll fix this, not sure why I left it empty.

Fabric.Authorization.API/Modules/UsersModule.cs
michaelvidal24: I'll add that verification.

Fabric.Authorization.API/Modules/UsersModule.cs
michaelvidal24: Same comment as above, since the parameters are in the route, we wouldn't route to this function if they are null.

Fabric.Authorization.API/Modules/UsersModule.cs
michaelvidal24: Good call. I'll make that change.

Fabric.Authorization.Domain/Services/UserService.cs
michaelvidal24: We've overriden ToString on Role, so it should get written out w/ the {role} variable.

Fabric.Authorization.Domain/Services/UserService.cs
michaelvidal24: I'll add those in.

Fabric.Authorization.Domain/Services/UserService.cs
bsmithvt: Oh I didn't notice the `{role}`.

PR #153, opened 2018-02-16
Update RolePermissionResolverService

michaelvidal24:
Now that we can associate users directly to roles, the RolePermissionResolverService needs to check if the role is associated to the user when deciding whether to process it or not.


Fabric.Authorization.Domain/Resolvers/Permissions/RolePermissionResolverService.cs
```diff
@@ -78,6 +78,16 @@ public RolePermissionResolverService(RoleService roleService)
             };
         }
 
+        private static bool BypassRoleEvaluation(Role role, List<string> groupNames, PermissionResolutionRequest permissionResolutionRequest)
+        {
```
bsmithvt: This was missing originally, but should `|| role.Permissions.Count == 0` also be in this condition?

michaelvidal24: That looks like a good performance optimization. I'll add it in.

PR #154, opened 2018-02-16
User/bsmithvt/securableitem dos

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Mappers/ClientMapperProfile.cs
```diff
@@ -14,8 +14,8 @@ public ClientMapperProfile()
                 .ForPath(x => x.ClientId, opt => opt.MapFrom(x => x.Id))
                 .ForMember(x => x.TopLevelSecurableItem, opt => opt.MapFrom(src => src.TopLevelSecurableItem))
                 .ForMember(x => x.Id, opt => opt.Ignore())
-                .ForMember(x => x.SecurableItemId, opt => opt.Ignore());
-         
+                .ForMember(x => x.SecurableItemId, opt => opt.Ignore())
+                .ForMember(x => x.TopLevelSecurableItem, opt => opt.MapFrom(src => src.TopLevelSecurableItem));
         }
```
kpaul: this looks like a duplicate line (line 15)

Fabric.Authorization.Domain/Services/SecurableItemService.cs
```diff
@@ -46,8 +63,13 @@ public SecurableItemService(IClientStore clientStore)
             item.Id = Guid.NewGuid();
             item.ClientOwner = clientId;
             var client = await _clientStore.Get(clientId);
-            var parentSecurableItem = GetSecurableItemById(client.TopLevelSecurableItem, itemId);
+            var parentSecurableItem = await GetSecurableItem(client.Id, itemId);
             CheckUniqueness(parentSecurableItem, item);
+
+            if (parentSecurableItem.Grain != item.Grain)
+            {
+                throw new BadRequestException<SecurableItem>("The SecurableItem child grain must match the parent SecurableItem's grain.");
```
michaelvidal24: Should the Domain know this is a BadRequest? Should we make this an exception that is domain specific, and not REST API specific?

bsmithvt: I'm not sure. The exception is a domain-level exception, and this and other services throw this exception along with `AlreadyExistsException` and `NotFoundException`. Did you want to rename `BadRequestException` to something else?

michaelvidal24: We can rename it in a future PR - just don't want HTTP concepts leaking into the domain.

PR #166, opened 2018-03-05
User/bsmithvt/fabric auth ui service

bsmithvt:
More work in progress on the Fabric Auth integration services.


Fabric.Authorization.AccessControl/src/app/modules/access-control/access-control.module.ts
```diff
@@ -5,15 +5,28 @@ import { AccessControlRoutingModule } from './access-control-routing.module';
 import { UserlistComponent } from './userlist/userlist.component';
 import { UseraddComponent } from './useradd/useradd.component';
 import { AuthserviceService } from '../../services/authservice.service';
+import { HttpClientModule } from '@angular/common/http';
```
kpaul: we may have to adjust our setup and not import this here. we need to sort out the interceptor stuff though i believe

PR #170, opened 2018-03-07
added member-list component with basic table display

kpaul:
wired up the containing application to be able to login and out with identity
adjusted the http interceptor a bit to get the headers working 

i will add the remaining stuff to the member-list view next i just wanted to get these changes committed as the number of changes i have to commit to master is getting large. i am also going to work on a service to inject into the http interceptor to get the access token per request.


Fabric.Authorization.AccessControl/src/app/modules/access-control/access-control-routing.module.ts
```diff
@@ -2,17 +2,17 @@ import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 
 
-import { UserlistComponent } from './userlist/userlist.component';
-import { UseraddComponent } from './useradd/useradd.component';
+import { MemberListComponent } from './member-list/member-list.component';
+import { MemberAddComponent } from './member-add/member-add.component';
 
 const routes: Routes = [
   {
```
bsmithvt: Does path need to be set here as well?

kpaul: Do you mean for the member list route? No it does not need to be able to because it is the default route for this router. The main router will route to this module and this router will load the member list component.

PR #171, opened 2018-03-07
added a config service for providing configuration data and functions to the access control module

kpaul:
the idea is that the app that installs the access control module will create a service that implements the functions and sets the properties we need to access the backend services (i.e. api urls, access token, client id). we can then import that service into any access control module services that require backend services.

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/member-list/member-list.component.ts
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/member-list/member-list.component.ts
@@ -4,8 +4,7 @@ import { Observable } from 'rxjs/Rx';
 import { AccessControlConfigService } from '../../../services/access-control-config.service';
 
 import { FabricAuthMemberSearchService } from '../../../services/fabric-auth-member-search.service';
-import { AuthMemberSearchRequest } from '../../../models/authMemberSearchRequest.model'
-import { AuthMemberSearchResult } from '../../../models/authMemberSearchResult.model';
+import { AuthMemberSearchRequest, AuthMemberSearchResult } from '../../../models'
 
 @Component({
   selector: 'app-member-list',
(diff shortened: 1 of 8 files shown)


Fabric.Authorization.AccessControl/src/app/modules/access-control/member-list/member-list.component.ts
```diff
@@ -1,9 +1,11 @@
 import { Component, OnInit } from '@angular/core';
 import { Observable } from 'rxjs/Rx';
 
+import { AccessControlConfigService } from '../../../services/access-control-config.service';
+
 import { FabricAuthMemberSearchService } from '../../../services/fabric-auth-member-search.service';
-import { AuthMemberSearchRequest } from '../../../models/authMemberSearchRequest';
-import { AuthMemberSearchResult } from '../../../models/authMemberSearchResult';
```
bsmithvt: Minor note: the `index.ts` file in the `models` (and other) directories allow `import` statements to be consolidated on a single line. The 2 lines below could become:

`import { AuthMemberSearchRequest, AuthMemberSearchResult } from '../../../models'`

kpaul: ok cool. i can switch that up.

Fabric.Authorization.AccessControl/src/app/services/global/client-access-control-config.service.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  constructor(private authService: AuthService) { }
+
+  clientId = 'fabric-angularsample';
+
+  getAccessToken(){
+     return this.authService.getUser()
+    .then(function(user){           
+    if(user){
+         return Promise.resolve(user.access_token);
```
benjanderson: you dont have to create another promise here, it will work just fine if you `return user.access_token;`

kpaul: ahh yeah you are correct. i will remove that extra promise

PR #176, opened 2018-03-09
User/bsmithvt/fabric auth ui service tests

bsmithvt:


Fabric.Authorization.AccessControl/src/app/services/fabric-auth-group.service.ts
```diff
@@ -32,6 +32,11 @@ export class FabricAuthGroupService extends FabricBaseService {
     }
   }
 
+  public getGroupUsers(groupName: string) : Observable<Group> {
+    return this.httpClient
+      .get<Group>(this.replaceGroupNameSegment(FabricAuthGroupService.groupUsersApiUrl, groupName));
```
michaelvidal24: Do we need to url encode the groupName, sometimes these will have `\` in them.

Fabric.Authorization.AccessControl/src/app/services/fabric-auth-user.service.ts
```diff
@@ -24,6 +25,15 @@ export class FabricAuthUserService extends FabricBaseService {
...
     }
+
+    if (!FabricAuthUserService.userGroupsApiUrl) {
+      FabricAuthUserService.userGroupsApiUrl = `${FabricAuthUserService.baseUserApiUrl}/{identityProvider}/{subjectId}/groups`;
+    }
+  }
+
+  public getUserGroups(identityProvider: string, subjectId: string) : Observable<string[]> {
+    return this.httpClient
+      .get<string[]>(this.replaceUserIdSegment(FabricAuthUserService.userGroupsApiUrl, identityProvider, subjectId));
```
michaelvidal24: Same comment as for groups, do we need to urlencode the subjectId as these may have `\` in them.

PR #181, opened 2018-03-13
added ability to add a member (group or user) via a directory search

kpaul:
added new endpoint to groups module for adding multiple roles to a group, updated code and tests accordingly

### DIFF
--- a/Fabric.Authorization.API/Modules/GroupsModule.cs
+++ b/Fabric.Authorization.API/Modules/GroupsModule.cs
@@ -194,6 +194,11 @@ namespace Fabric.Authorization.API.Modules
             }
         }
 
+        private async Task<dynamic> AddRolesToGroup()
+        {
+            
+        }
+
         private async Task<dynamic> DeleteRoleFromGroup()
         {
             try
--- a/Fabric.Authorization.API/Modules/GroupsModule.cs
+++ b/Fabric.Authorization.API/Modules/GroupsModule.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Fabric.Authorization.API.Configuration;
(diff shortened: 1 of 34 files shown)


Fabric.Authorization.API/Modules/GroupsModule.cs
```diff
@@ -165,32 +178,36 @@ public class GroupsModule : FabricModule<Group>
...
-                if (groupRoleRequest.Id == null && groupRoleRequest.RoleId == null)
-                {
-                    return CreateFailureResponse("Role ID is required.", HttpStatusCode.BadRequest);
-                }
+                return errorResponse;
+            }
 
-                var roleId = groupRoleRequest.Id ?? groupRoleRequest.RoleId;
-                var group = await _groupService.AddRoleToGroup(groupRoleRequest.GroupName, roleId.Value);
-                return CreateSuccessfulPostResponse(group.ToGroupRoleApiModel());
```
bsmithvt: Not needed for this PR, but maybe `CheckWriteAccess` can be changed in the future to accept a collection of tuples containing grain/securableItem pairs and it can loop over them internally.

Fabric.Authorization.AccessControl/src/app/models/fabricPrincipal.model.ts
```diff
@@ -6,5 +6,7 @@ export class FabricPrincipal {
     public lastName: string;
     public principalType: string;
```
bsmithvt: Very minor - if `middleName` is blank, there will be 2 spaces between `firstName` and `lastName` in the UI.

Fabric.Authorization.AccessControl/src/app/modules/access-control/member-add/member-add.component.ts
```diff
@@ -7,9 +17,76 @@ import { Component, OnInit } from '@angular/core';
...
+    } else{
+      this.selectedRoles = this.selectedRoles.filter(r => r.name !== role.name);
+    }
+  }
+
+  roleIsSelected(role: Role){
+    return this.selectedRoles.filter(r => r.name === role.name).length > 0;
+  }
+
+  getRoles(){
```
bsmithvt: This is passing in a `clientId` and not a `securableItem` for the second parameter.

Fabric.Authorization.AccessControl/src/app/services/fabric-auth-group.service.ts
```diff
@@ -67,11 +67,21 @@ export class FabricAuthGroupService extends FabricBaseService {
       .post<Group>(this.replaceGroupNameSegment(FabricAuthGroupService.groupRolesApiUrl, groupName), role);
   }
```
bsmithvt: I think we need to remove the method above (`addRoleToGroup`) since that's no longer supported.

Fabric.Authorization.AccessControl/src/app/models/fabricPrincipal.model.ts
kpaul: good catch i actually meant to remove this

Fabric.Authorization.AccessControl/src/app/services/fabric-auth-group.service.ts
kpaul: yeah, i will do that

Fabric.Authorization.API/Modules/GroupsModule.cs
kpaul: not a bad idea

Fabric.Authorization.AccessControl/src/app/modules/access-control/member-add/member-add.component.ts
kpaul: ahh yeah. they are the same value but i will switch it. a lot of this code is probably going away anyway once the data governance team does the real ui

PR #189, opened 2018-03-15
added component to add a custom group and associate with users and roles

kpaul:
updated member search response to include the role model for members instead of just the role id. 

fixed a minor bug in the groups metadata module


michaelvidal24:
One other comment/question - are we planning to enhance the members endpoint to take in a grain/securable so we can get users tied to shared grains?

bsmithvt:
@michaelvidal24 , yes, that's the task I'm going to pick up next.

PR #190, opened 2018-03-16
enhanced DELETE /groups/{groupName}/roles endpoint to support bulk de…

bsmithvt:
…letes


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
```diff
@@ -189,19 +189,35 @@ public SqlServerGroupStore(IAuthorizationDbContext authorizationDbContext)
...
-                await _authorizationDbContext.SaveChangesAsync();
+                throw new NotFoundException<Role>($"No role mappings found for group {group.Name} with the supplied role IDs");
+            }
+
+            foreach (var groupRole in groupRolesToRemove)
+            {
+                // remove the role from the domain model
+                var roleToRemove = group.Roles.FirstOrDefault(r => r.Id == groupRole.RoleId);
+
+                if (roleToRemove == null)
```
michaelvidal24: Do we want to build up these extensions so we can return a single error response if multiple roles don't exist on the group? Not suggesting we have to, just wondering what your thoughts are here.

bsmithvt: Yes, good idea.

PR #196, opened 2018-03-22
add component for editing members

kpaul:
there was also a decent amount of api changes necessary to support adding and removing roles from users and groups


Fabric.Authorization.API/Modules/GroupsModule.cs
```diff
@@ -75,6 +75,11 @@ public class GroupsModule : FabricModule<Group>
                 null,
                 "GetRolesFromGroup");
 
+            Get("/{groupName}/{grain}/{securableItem}/roles",
```
michaelvidal24: Is it possible to collapse this endpoint with the GET /groups/{groupName}/roles endpoint? Possibly by reading qs params for grain and securableItem?

kpaul: it might be possible. my concern is that we are getting the roles for the client in that code and then doing an intersect with the roles for the group. would it be possible that the group may have a role that we want returned that is not tied to the client being searched?

PR #202, opened 2018-03-28
AccessControl: Make Identity server load from environment settings

joeskeen:
... also allow for 'user' environment which is excluded from source control

In order to test, we have set up Fabric Identity and Auth on the AtlasDemo server our team uses.  This change allows developers to create their own environment file, `environment.user.ts`, which contains user-specific settings to use for testing without having those changes be accidentally committed to source control.  For example, my `environment.user.ts` file looks like this:

```ts
import { Environment, defaultEnvironment } from './environment.default';

export const environment = Object.assign<Environment, Partial<Environment>>(defaultEnvironment, {
  fabricAuthApiUri: 'https://atlasdemo.hqcatalyst.local/Authorization',
  fabricIdentityApiUri: 'https://atlasdemo.hqcatalyst.local/Identity',
  fabricExternalIdPSearchApiUri: 'https://atlasdemo.hqcatalyst.local/IdentityProviderSearchService',
});

```

I can build / serve the project with my custom environment by passing `--env=user` to the Angular CLI.

Also, as part of this, I made a default settings file, so each environment only needs to specify which default settings they are overriding, preventing environment files from getting out of sync when new settings are added, renamed, or removed (TS will alert you in those cases).


kpaul:
Good idea!

PR #210, opened 2018-04-03
updates to run karma tests with phantomjs

kpaul:


Fabric.Authorization.AccessControl/karma.conf.js
```diff
@@ -27,7 +28,7 @@ module.exports = function (config) {
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: true,
-    browsers: ['Chrome'],
+    browsers: [ 'Chrome', 'PhantomJS' ],
```
joeskeen: PhantomJS has been deprecated in favor of HeadlessChrome.  Is there a particular reason you want to use PhantomJS?

kpaul: No particular reason, and I would be happy to avoid using it. When I was searching around for running karma tests in VSTS builds everything I saw referenced PhantomJS. I will look into HeadlessChrome

joeskeen: I believe we have gotten it working in VSTS Hosted agents for Atlas.  Let me check out what we're doing and I we can talk.

kpaul:
merging with changes to use chrome headless. i need to update the build to install some dependencies to get it loaded on linux and i want to be able to kick off builds manually to test which i cannot do from here.

PR #213, opened 2018-04-03
add headless chrome flags

kpaul:


Fabric.Authorization.AccessControl/karma.conf.js
```diff
@@ -30,6 +30,18 @@ module.exports = function (config) {
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: [ 'Chrome', 'ChromeHeadless'],
-    singleRun: false
+    singleRun: false,
```
joeskeen: you are passing single run on the script that runs in CI, right?

kpaul: yes

PR #215, opened 2018-04-03
combine member add and edit

benjanderson:


Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.html
```diff
@@ -3,12 +3,12 @@ <h3 class="h3">
     <span>Name (as it appears in directory)</span>
   </h3>
   <div hcInputIcon>
-    <input hc-input class="hc-input" (keyup)="searchText.next($event.target.value)">
+    <input hc-input class="hc-input" (keyup)="searchTextSubject.next($event.target.value)" [(ngModel)]="subjectId" [readOnly]="editMode">
```
joeskeen: Would it better to not have it be an input, perhaps a title or label instead when editing?  The presence of a disabled input box can make the user think they either don't have permission to the box or that they need to do something to enable it.  As you will never have it enabled for an existing user record, IMHO it makes sense to not have an input box at all when `editMode`

Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  private bindExistingRoles(subjectId: string, principalType: string): Observable<any> {
+    if (!this.subjectId) {
+      return Observable.of(undefined);
+    }
+    const roleObservable = principalType === 'user' ?
+      this.userService.getUserRoles(this.configService.identityProvider, subjectId) :
+      this.groupService.getGroupRoles(subjectId, this.configService.grain, this.configService.securableItem);
+
+    return roleObservable.map((existingRoles: IRole[]) => {
```
joeskeen: when calling `Observable.map`, shouldn't you be returning something?

Fabric.Authorization.AccessControl/src/app/services/fabric-auth-user.service.ts
```diff
@@ -69,6 +69,10 @@ export class FabricAuthUserService extends FabricBaseService {
     subjectId: string,
     roles: IRole[]
   ): Observable<IUser> {
+    if (!roles || roles.length === 0) {
+      return Observable.of(undefined);
```
joeskeen: would it be better to return an observable of `undefined`, or an empty observable?  It seems to me that if you do that, the subscriber will have to check the value for `undefined`, right?

Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.ts
benjanderson: good catch!

Fabric.Authorization.AccessControl/src/app/services/fabric-auth-user.service.ts
benjanderson: done

Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.html
joeskeen: TODO: get feedback from others on what we should do about this

Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.ts
joeskeen: I think this will return an Observable<Observable<IRole>>.  Is this what you intended?

PR #231, opened 2018-04-12
Adding upgrade functionality to move EDWAdmin roles to Fabric.Auth roles

petermcclanahan:
At the end of the script, I added a query to get the users and groups from EDWAdmin, then iterated over the list and called the already existing `Add-AccountToDosAdminRole` which does all the error checking and adds it to the right table based on whether it is a group or a user.

I also fixed the formatting according to my VS Code preferences... That's a lot of changes and I'm happy to undo them if you prefer. Just let me know. I'll highlight the changes in the PR

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
@@ -2,35 +2,32 @@
 # Install_Authorization_Windows.ps1
 #
 param([switch]$noDiscoveryService)
-function Get-DiscoveryServiceUrl()
-{
+function Get-DiscoveryServiceUrl() {
     return "https://$env:computername.$($env:userdnsdomain.tolower())/DiscoveryService"
 }
 
-function Get-IdentityServiceUrl()
-{
+function Get-IdentityServiceUrl() {
     return "https://$env:computername.$($env:userdnsdomain.tolower())/Identity"
 }
 
-function Get-Headers($accessToken){
+function Get-Headers($accessToken) {
     $headers = @{"Accept" = "application/json"}
-    if($accessToken){
+    if ($accessToken) {
         $headers.Add("Authorization", "Bearer $accessToken")
     }
     return $headers
 }
 
-function Add-DiscoveryRegistration($discoveryUrl, $serviceUrl, $credential)
-{	
+function Add-DiscoveryRegistration($discoveryUrl, $serviceUrl, $credential) {	
     $registrationBody = @{
-        ServiceName = "AuthorizationService"
-        Version = 1
-        ServiceUrl = $serviceUrl
+        ServiceName   = "AuthorizationService"
+        Version       = 1
+        ServiceUrl    = $serviceUrl
         DiscoveryType = "Service"
-        IsHidden = $true
-        FriendlyName = "Fabric.Authorization"
-        Description = "The Fabric.Authorization service provides centralized authorization across the Fabric ecosystem."
-        BuildNumber = "1.1.2017120101"
+        IsHidden      = $true
+        FriendlyName  = "Fabric.Authorization"
+        Description   = "The Fabric.Authorization service provides centralized authorization across the Fabric ecosystem."
+        BuildNumber   = "1.1.2017120101"
     }
 
 	$url = "$discoveryUrl/v1/Services"
@@ -49,9 +46,8 @@ function Add-DiscoveryRegistration($discoveryUrl, $serviceUrl, $credential)
 	}
 }
 
-function Add-DatabaseLogin($userName, $connString)
-{
-	$query = "USE master
+function Add-DatabaseLogin($userName, $connString) {
+    $query = "USE master
 			If Not exists (SELECT * FROM sys.server_principals
 				WHERE sid = suser_sid(@userName))
 			BEGIN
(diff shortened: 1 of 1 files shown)


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -248,19 +244,38 @@ function Get-ErrorFromResponse($response)
...
-	$url = "$authorizationUrl/_monitor/shallow"
-	Invoke-RestMethod -Method Get -Uri $url
+function Invoke-MonitorShallow($authorizationUrl) {
+    $url = "$authorizationUrl/_monitor/shallow"
+    Invoke-RestMethod -Method Get -Uri $url
 }
 
-if(!(Test-Path .\Fabric-Install-Utilities.psm1)){
-	Invoke-WebRequest -Uri https://raw.githubusercontent.com/HealthCatalyst/InstallScripts/master/common/Fabric-Install-Utilities.psm1 -OutFile Fabric-Install-Utilities.psm1
+function Get-EdwAdminUsersAndGroups() {
```
petermcclanahan: Added these two functions

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -688,21 +712,27 @@ Add-AccountToDosAdminRole -accountName $adminAccount -domain $currentUserDomain
...
+if ($authorizationServiceUrl) { Add-InstallationSetting "authorization" "authorizationService" "$authorizationServiceUrl" | Out-Null}
+if ($authorizationServiceUrl) { Add-InstallationSetting "common" "authorizationService" "$authorizationServiceUrl" | Out-Null}
+if ($iisUser) { Add-InstallationSetting "authorization" "iisUser" "$iisUser" | Out-Null}
+if ($siteName) {Add-InstallationSetting "authorization" "siteName" "$siteName" | Out-Null}
+if ($adminAccount) {Add-InstallationSetting "authorization" "adminAccount" "$adminAccount" | Out-Null}
 
 
 Invoke-MonitorShallow "$hostUrl/$appName"
 
+Write-Host "Upgrading all the users/groups with an 'EDW Admin' role to also have the dosadmin Fabric.Auth role"
```
petermcclanahan: Added this part from here on down

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -248,19 +244,38 @@ function Get-ErrorFromResponse($response)
...
-	Invoke-RestMethod -Method Get -Uri $url
+function Invoke-MonitorShallow($authorizationUrl) {
+    $url = "$authorizationUrl/_monitor/shallow"
+    Invoke-RestMethod -Method Get -Uri $url
 }
 
-if(!(Test-Path .\Fabric-Install-Utilities.psm1)){
-	Invoke-WebRequest -Uri https://raw.githubusercontent.com/HealthCatalyst/InstallScripts/master/common/Fabric-Install-Utilities.psm1 -OutFile Fabric-Install-Utilities.psm1
+function Get-EdwAdminUsersAndGroups() {
+    $results = Invoke-Sqlcmd -Query "
```
michaelvidal24: Since this query will run on the web app machine, do we need to specify the ServerInstance to run against? I think this function also requires the sqlserver module, which has to be installed. We have a Invoke-Sql method in the Fabric-Install-Utilities.psm1 which we already import that doesn't have these pre-reqs. For a usage example see line 151 of this script.

petermcclanahan: Yeah, you're right.

I tried using the Invoke-Sql method, but I was having a hard time getting results back (and I didn't see an example where that was happening). Maybe it's worth looking at again and possibly updating that...

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -248,19 +244,38 @@ function Get-ErrorFromResponse($response)
...
+            ON ir.IdentityID = i.IdentityID
+        WHERE RoleNM = 'EDW Admin'";
+
+    return $results;
+}
+
+function Add-UsersAndGroupsToDosAdminRole($edwAdminUsersAndGroups, $domain, $connString) {
+
+    foreach ($edwAdmin in $edwAdminUsersAndGroups) {
+        Add-AccountToDosAdminRole -accountName $edwAdmin.IdentityNM -domain $domain -authorizationServiceUrl "$authorizationServiceUrl/v1" -accessToken $accessToken -connString $authorizationDbConnStr
```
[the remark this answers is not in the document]
petermcclanahan: Yes! Good catches. I meant to do that, but I guess I missed it.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
michaelvidal24: Let me know if you can't get the Invoke-Sql function to work and I can help take a look at it.

michaelvidal24: Looking deeper, the Invoke-Sql function does an ExecuteNonQuery, so it won't return a result set. We can look into adding another function that does return a result set, I'm just concerned about introducing a dependency on the sqlserver powershell module.

petermcclanahan: Yeah, I noticed that, too. Let me know what you think of this approach. I considered making a generic method in the Fabric-Install-Utilities module, but it can get pretty complex to pass in what columns you want returned from the query you pass in, and rather than over-engineering I decided to just put it inside the `Get-EdwAdminUsersAndGroups` function. It uses a connection string now, and not `Invoke-Sqlcmd`,.

PR #233, opened 2018-04-20
User/bsmithvt/sqlserver serilog auditing

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerUserStore.cs
```diff
@@ -134,28 +139,33 @@ public SqlServerUserStore(IAuthorizationDbContext authorizationDbContext)
...
+                };
+
+                AuthorizationDbContext.RoleUsers.Add(roleUser);
+                roleUsers.Add(roleUser);
             }
 
-            await _authorizationDbContext.SaveChangesAsync();
-            return user;            
+            await AuthorizationDbContext.SaveChangesAsync();
+            await EventService.RaiseEventAsync(new EntityBatchAuditEvent<RoleUser>(EventTypes.EntityCreatedEvent, roleUsers));
```
michaelvidal24: In this case should we just use EntityAuditEvent and audit the User domain object, instead of the RoleUser entity object? I think we would get a more holistic picture of what changed if we audit the domain object changes. I think this would apply to GroupRoles and GroupUsers too. Thoughts?

bsmithvt: I actually had it implemented that way at first, but wasn't sure if the log would contain what actually changed (the many to many entities) in these scenarios. I think it makes sense to change it back like you suggested.

bsmithvt: @michaelvidal24, I'm thinking of adding a new event type called `ChildEntityCreated` and `ChildEntityDeleted` for these scenarios so we have some extra context. Do you think this makes sense?

michaelvidal24: As opposed to caputuring the whole parent entity w/ children? Or in addition to? I like the idea if it is in addition to capturing the whole domain entity. We'll just have to make sure it doesn't hurt performance, which we'll know that by looking at the results of the peformance tests in the release pipeline.

bsmithvt: I meant in addition to capturing the whole domain entity. Otherwise adding roles to a group will appear as a "group created" event.

michaelvidal24: Makes sense to me.

PR #238, opened 2018-04-24
updated group name input on add custom group page to be a typeahead

kpaul:
when a user starts to enter a custom group a search will be done against existing custom groups and those with a name that starts with what the user enters will be returned. the user can then select an existing group or enter a new one.

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
@@ -8,6 +8,7 @@
         <div class="form-errors" *ngIf="groupNameinvalid" [innerText]="groupNameError"></div>
     </div>
     <div>
+        <div *ngIf="customGroups.length > 0">Select an existing group:</div>
         <ul *ngFor="let group of customGroups">
             <li>
                 <hc-checkbox [id]="group.id" [(ngModel)]="group.selected" (click)="customGroupSelected(group)">{{group.groupName}}</hc-checkbox>
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
@@ -4,7 +4,15 @@
     </h3>
     <div class="input-combo">
         <input hc-input class="hc-input" required [(ngModel)]="groupName" [readOnly]="editMode" [highlight]="groupNameinvalid" (keyup)="groupNameSubject.next($event.target.value)">
+        <i [class]="searchingGroup ? 'btn-icon fa fa-spinner fa-pulse fa-fw' : 'btn-icon fa fa-search'"></i>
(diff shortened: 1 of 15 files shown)


Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -202,5 +202,20 @@ public class GroupService
...
         }
+
+        public async Task<IEnumerable<Group>> GetGroups(string name, string type)
+        {
+            if (!string.IsNullOrEmpty(type))
+            {
+                type = type.ToLower();
+                if (!type.Equals("custom") && !type.Equals("directory"))
+                {
+                    throw new BadRequestException<Group>("Invalid type provided. If provided valid values are custom or directory");
```
michaelvidal24: Since this is in the domain layer, should we change this to an InvaildOperationException? "Bad Request" is an HTTP type of concept which I don't think we should have in the domain layer. The Web layer can catch the InvalidOperationException and create a bad request response.

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -202,5 +202,20 @@ public class GroupService
 
             return await _groupStore.DeleteUserFromGroup(group, user);
         }
+
+        public async Task<IEnumerable<Group>> GetGroups(string name, string type)
+        {
+            if (!string.IsNullOrEmpty(type))
+            {
+                type = type.ToLower();
```
bsmithvt: We have a `CustomSource` constant in `GroupConstants`. Perhaps we can use that here and and another `DirectorySource` constant to that class for use everywhere?

kpaul: good call, I forgot about that.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
```diff
@@ -3,9 +3,18 @@ <h3 class="h3">
...
     <div class="input-combo">
-        <input hc-input class="hc-input" required [(ngModel)]="groupName" [readOnly]="editMode" [highlight]="groupNameinvalid" (keyup)="groupNameSubject.next($event.target.value)">
+        <input hc-input class="hc-input" required [(ngModel)]="groupName" [readOnly]="editMode" [highlight]="groupNameinvalid" (keyup)="groupNameSubject.next($event.target.value)" placeholder="enter group name">
+        <i [class]="searchingGroup ? 'btn-icon fa fa-spinner fa-pulse fa-fw' : 'btn-icon fa fa-search'"></i>
         <div class="form-errors" *ngIf="groupNameinvalid" [innerText]="groupNameError"></div>
     </div>
+    <div class="scroll-list">
+        <div *ngIf="customGroups.length > 0"><span>Select an existing group:</span></div>
+        <ul>
+            <li *ngFor="let group of customGroups">
```
benjanderson: not using the scroll-table will make it not consistent with the member page. I am ok converting over that component to using <li> but I would prefer them to be consistent.

kpaul: Ok sure, I can change the scroll list to be a scroll-table

PR #254, opened 2018-05-29
1) updated README.md, 2) updated default IdP search service URI

bsmithvt:


Fabric.Authorization.AccessControl/README.md
```diff
@@ -1,13 +1,93 @@
 # Fabric.Authorization.AccessControl
 An Angular 5 routed module for managing access in Fabric.Authorization
 
+# Documentation
+[Requirements](https://healthcarequalitycatalyst.sharepoint.com/:o:/r/productdev/_layouts/15/WopiFrame.aspx?sourcedoc={32fa060b-915f-4641-a046-5f538a42f52b}&action=edit&wdLOR=&wd=target%28%2F%2FSprint%20Notes%2FFabricApp.one%7Cc10158a6-70a1-436e-b1be-cfe0af705094%2FAccess%20control%20portal%20Phase%201%7C470b5cf5-9bfa-4f42-b83d-78fd94fa3d17%2F%29)
```
ryanorbaker: not sure what option we have, but I'm not sure I like the idea of a catalyst sharepoint URL in a github (public) readme.

bsmithvt: I thought this specific README document was for internal developers only who want to set up their local machines to run the sample application, but I could be wrong. Is Cashmere something that will be open to the public? Because the document references that as well.

PR #255, opened 2018-05-30
Fixed an issue where we weren't checking to see if the user had permi…

michaelvidal24:
…ssions to manage roles tied to groups.

### DIFF
--- a/Fabric.Authorization.API/Modules/GroupsModule.cs
+++ b/Fabric.Authorization.API/Modules/GroupsModule.cs
@@ -267,12 +267,9 @@ namespace Fabric.Authorization.API.Modules
                 return errorResponse;
             }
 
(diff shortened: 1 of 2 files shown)


Fabric.Authorization.API/Modules/GroupsModule.cs
```diff
@@ -439,5 +440,20 @@ private Negotiator ValidateGroupUserRequest(GroupUserRequest groupUserRequest)
             }
             return null;
         }
+
```
bsmithvt: Can this method also be used in the `AddRolesToGroup` function? Perhaps it would need a minor refactor to accomplish that.

michaelvidal24: We could potentially extract a function that iterates through roles and checks write access. I'll refactor that.

PR #259, opened 2018-06-04
No space when storing authorization client secret

ctediz:
Had to cast and trim, trim/cast alone were not enough

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
@@ -633,7 +633,7 @@ $body = @'
 '@
 
 Write-Host "Registering Fabric.Authorization Client."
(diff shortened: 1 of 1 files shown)


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -633,7 +633,7 @@ $body = @'
 '@
 
 Write-Host "Registering Fabric.Authorization Client."
-$authorizationClientSecret = Add-ClientRegistration -authUrl $identityServerUrl -body $body -accessToken $accessToken
+$authorizationClientSecret = ([string](Add-ClientRegistration -authUrl $identityServerUrl -body $body -accessToken $accessToken)).Trim()
```
michaelvidal24: Do we also need to cast to a string and then trim when we are adding the Api registration above in line 617?

PR #261, opened 2018-06-11
Migrate all users from EDWAdmin to DosAdmin.

hc-jordenlowe:
Adapted code from https://github.com/HealthCatalyst/Fabric.Authorization/commit/823c2e946514a5915c1ed7ebec6ed883c1276406#diff-e81f46f8c24fc699274b63f0d42b2674

Created a function to get all users from EDWAdmin.  The neat thing is that this should only be users.  Groups should not have a role 'EDW Admin'.  The function could be reused, so I left it with a generic 'Get-EdwAdminUsersAndGroups', but could rename it to 'Get-EdwAdminUsers'.

After getting the users, we call 'Add-ListOfUsersToDosAdminRole' which is very similar to 'Add-AccountToDosAdminRole', but we do not check AD to verify that there is a user or group of that name.  Moreover it processes a list.  The for loop will swallow exceptions, but will print the error for the user in the script.  If the user exists, we will catch the 409 and will update the user saying we already of that user.

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
@@ -247,6 +247,67 @@ function Invoke-MonitorShallow($authorizationUrl) {
     Invoke-RestMethod -Method Get -Uri $url
 }
 
+function Get-EdwAdminUsersAndGroups($connectionString) {	
+    $connection = New-Object System.Data.SqlClient.SqlConnection($connectionString)	
+    $sql = "SELECT i.IdentityID, i.IdentityNM, r.RoleNM	
+            FROM [CatalystAdmin].[RoleBASE] r	
+            INNER JOIN [CatalystAdmin].[IdentityRoleBASE] ir	
+                ON r.RoleID = ir.RoleID	
+            INNER JOIN [CatalystAdmin].[IdentityBASE] i	
+                ON ir.IdentityID = i.IdentityID	
+            WHERE RoleNM = 'EDW Admin'";	
+    $command = New-Object System.Data.SqlClient.SqlCommand($sql, $connection)	
+    
+    $usersAndGroups = @();	
+    try {	
+        $connection.Open()    	
+        $reader = $command.ExecuteReader()	
+        while ($reader.Read()) {	
+            $usersAndGroups += $reader['IdentityNM']	
+        }	
+        $connection.Close()        	
+    
+    }	
+    catch [System.Data.SqlClient.SqlException] {	
+        Write-Error "An error ocurred while executing the command. Please ensure the connection string is correct and the metadata database has been setup. Connection String: $($connectionString). Error $($_.Exception.Message)"  -ErrorAction Stop	
+    }    	
+    
+    return $usersAndGroups;	
+}	
+    	
+function Add-ListOfUsersToDosAdminRole($edwAdminUsers, $connString, $authorizationServiceUrl, $accessToken) {	   
+    # Get the role once, should be same for every user.
+    $role = Get-Role -name "dosadmin" -grain "dos" -securableItem "datamarts" -authorizationServiceUrl $authorizationServiceUrl -accessToken $accessToken
+
+    # For each user, loop and try to add the user to the API.  
+    # Do not validate it to AD like the Add-AccountToDosAdminRole function.
+    foreach ($edwAdmin in $edwAdminUsers) {	
+        if ([string]::IsNullOrWhiteSpace($edwAdmin)) {
+            continue
+        }
+
+        try {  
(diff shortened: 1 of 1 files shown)


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -247,6 +247,63 @@ function Invoke-MonitorShallow($authorizationUrl) {
...
+        $connection.Open()    	
+        $reader = $command.ExecuteReader()	
+        while ($reader.Read()) {	
+            $usersAndGroups += $reader['IdentityNM']	
+        }	
+        $connection.Close()        	
+    
+    }	
+    catch [System.Data.SqlClient.SqlException] {	
+        Write-Error "An error ocurred while executing the command. Please ensure the connection string is correct and the metadata database has been setup. Connection String: $($connectionString). Error $($_.Exception.Message)"  -ErrorAction Stop
```
hc-jordenlowe: If we cannot connect to the EDWAdmin database, this will stop the script.  Is that a good thing?  This executes near the bottom, so this wouldnt stop the process.  If we want to make this a warning, we could remove the ErrorAction Stop part.

ctediz:
Wasn't something similar removed is it coming back now?

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
michaelvidal24: I think it's probably safe to keep this. If we can't connect to the DB to get the users, there is no need to run any steps beyond the query anyway.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -709,4 +766,12 @@ if ($adminAccount) {Add-InstallationSetting "authorization" "adminAccount" "$adm
 
 Invoke-MonitorShallow "$authorizationServiceUrl"
 
+Write-Host "Upgrading all the users with an 'EDW Admin' role to also have the dosadmin Fabric.Auth role"	+Read-Host -Prompt "Installation complete, press Enter to exit"
```
michaelvidal24: Do we need to remove the `Read-Host` here?

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -247,6 +247,63 @@ function Invoke-MonitorShallow($authorizationUrl) {
...
+    
+    }	
+    catch [System.Data.SqlClient.SqlException] {	
+        Write-Error "An error ocurred while executing the command. Please ensure the connection string is correct and the metadata database has been setup. Connection String: $($connectionString). Error $($_.Exception.Message)"  -ErrorAction Stop	
+    }    	
+    
+    return $usersAndGroups;	
+}	
+    	
+function Add-ListOfUsersToDosAdminRole($edwAdminUsers, $domain, $connString, $authorizationServiceUrl, $accessToken) {
```
michaelvidal24: Since we aren't validating users against the domain, should we drop the $domain parameter?

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -709,4 +766,12 @@ if ($adminAccount) {Add-InstallationSetting "authorization" "adminAccount" "$adm
 
 Invoke-MonitorShallow "$authorizationServiceUrl"
 
+Write-Host "Upgrading all the users with an 'EDW Admin' role to also have the dosadmin Fabric.Auth role"	+Read-Host -Prompt "Installation complete, press Enter to exit"
+# There are no groups that will have 'EDW Admin'.  Should be only users.
+# For more information check out PBI 143708
+$edwAdminUsers = Get-EdwAdminUsersAndGroups -connectionString $metadataConnStr	
+Write-Host "There are $($edwAdminUsers.Count) users with this role"	
+Write-Host ""	
+Add-ListOfUsersToDosAdminRole -edwAdminUsers $edwAdminUsers -domain $currentUserDomain -connString $authorizationDbConnStr -authorizationServiceUrl "$authorizationServiceUrl/v1" -accessToken $accessToken
```
michaelvidal24: I think we can drop the `-domain` parameter.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
hc-jordenlowe: oh, i guess this was something that got thrown on that line.  I will remove it.  It is a duplicate.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
hc-jordenlowe: yep, we can remove that.

PR #266, opened 2018-06-18
Adding first version of Authorization client

hc-jordenlowe:
I figured we would want this out so we can work on the second PBI to this.  I think this is all the code that is required, but if i forgot anything, let me know. 

Later we will fix up the nuget, release pipeline and etc.


hc-jordenlowe:
thats no problem.  I didnt look at that close enough.

ryanorbaker:
should we talk about multiple sln files in the same directory, using a different subset of sub directories?  I worry about unintended side effects of shared sub projects.  Not sure if there is a better way.  Honesty, just checking.

michaelvidal24:
I agree, I don't think we need a separate solution for the Client SDK, simply having it as a separate project should be sufficient.

PR #269, opened 2018-06-19
Register Authorization BuildNumber with discovery

ctediz:
Remove Add-DiscoveryRegistration (moved into fabric utils)
Move Add-DiscoveryRegistration to after site deploy step so dll exists

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
@@ -626,12 +626,16 @@ if (!($noDiscoveryService)) {
     Write-Host ""
     Write-Host "Registering with Discovery Service."
     Write-Host ""
-    $buildVersion = [System.Diagnostics.FileVersionInfo]::GetVersionInfo("$appDirectory\Fabric.Authorization.API.dll").FileVersion
-    $serviceName = "AuthorizationService"
-    $serviceVersion = 1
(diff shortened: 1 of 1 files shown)


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -647,18 +626,15 @@ if (!($noDiscoveryService)) {
     Write-Host ""
     Write-Host "Registering with Discovery Service."
     Write-Host ""
-    Add-DiscoveryRegistration $discoveryServiceUrl "$authorizationServiceUrl/v1" $credential
+    $buildVersion = [System.Diagnostics.FileVersionInfo]::GetVersionInfo("$appDirectory\Fabric.Authorization.API.dll").FileVersion
+    $serviceName = "AuthorizationService"
+    $serviceVersion = 1
+    $friendlyName = "Fabric.Authorization"
+    $description = "The Fabric.Authorization service provides centralized authorization across the Fabric ecosystem."
+    Add-DiscoveryRegistration $discoveryServiceUrl "$authorizationServiceUrl/v1" $credential, $buildVersion, $serviceName, $serviceVersion, $friendlyName, $description
```
michaelvidal24: Same comment as in the other repo, it might be better to encapsulate the body of the post into its own object so we can reduce the number of parameters that need to be passed to this function.

PR #271, opened 2018-06-20
Changed Client to use 1.1 and more unit tests

hc-jordenlowe:
Should be 4 unit tests:

1. One for happy path, parses an API User
2. One to handle client errors.
3. One to handle server errors (like say if they are not of a json form)
4. One to handle null access tokens.

This also changed up the client for the better.  On top of that moved all projects to be .net 1.x

### DIFF
--- a/Fabric.Authorization.Client/AssemblyInfo.cs
+++ b/Fabric.Authorization.Client/AssemblyInfo.cs
@@ -1,0 +1,10 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Fabric.Authorization.Client.UnitTests")]
+
+namespace Fabric.Authorization.Client
+{
+    public class AssemblyInfo
+    {
+    }
+}
--- a/Fabric.Authorization.Client/Routes/ClientRoute.cs
+++ b/Fabric.Authorization.Client/Routes/ClientRoute.cs
@@ -1,8 +1,4 @@
-﻿using System.Runtime.CompilerServices;
-
-[assembly: InternalsVisibleTo("Fabric.Authorization.Client.UnitTests")]
-
-namespace Fabric.Authorization.Client.Routes
+﻿namespace Fabric.Authorization.Client.Routes
 {
     internal class ClientRoute : BaseRoute
     {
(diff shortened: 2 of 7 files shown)


hc-jordenlowe:
Just added nuspec, this will be used when we create the nuget pack in the CI pipeline

Fabric.Authorization.Client/scripts/Fabric.Authorization.Client.nuspec
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
michaelvidal24: Did we end up needing a nuspec file? I thought I remembered something about .net standard project using the .csproj file to figure out what needs to be packed. I could be wrong though.

Fabric.Authorization.Client.UnitTests/AuthorizationClientTests.cs
```diff
@@ -1,35 +1,145 @@
+using System;
```
bsmithvt: What is everyone's thoughts on establishing a naming convention for unit/integration test method names? Below is the convention I've used so far for Identity/Auth (and other companies in the past) that seems to work well:

http://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html

hc-jordenlowe: I can rename these to something of the nature:

<Function name>_<What we are testing>_<Expected result>

That naming convention works fine, but may be long.  Depends on the test runner used, if you need all of that information.  But I am fine with that format.

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -15,13 +16,29 @@ public AuthorizationClient(HttpClient client)
             this.client = client;
         }
 
-        public async Task<UserApiModel> GetPermissionsForCurrentUser()
+        public async Task<UserApiModel> GetPermissionsForCurrentUser(string accessToken)
         {
+            CheckAccessToken(accessToken);
+
```
michaelvidal24: I missed this on the original code review, but does HttpRequestMessage need to be wrapping in a using? I think it implements IDisposable.

PR #273, opened 2018-06-21
added route builders

bsmithvt:


bsmithvt:
Do we want to make these classes `public` to allow developers to use these classes for constructing routes for their `HttpClient`?

michaelvidal24:
I would vote to keep the classes internal for now. I'd rather have a smaller, more opinionated API surface to support than to provide a ton of functionality/flexibility. It will allow us the flexibility to refactor internal concerns more easily, and also make it easier for people to learn the API.

bsmithvt:
@michaelvidal24 , I agree.

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -15,13 +18,240 @@ public AuthorizationClient(HttpClient client)
...
+        }
+
+        public async Task AddPermissionsToUser(string identityProvider, string subjectId, List<PermissionApiModel> permissionModels)
+        {
+            var message = new HttpRequestMessage(HttpMethod.Post, new UserRouteBuilder().IdentityProvider(identityProvider).SubjectId(subjectId).UserPermissionsRoute)
+            {
+                Content = new StringContent(JsonConvert.SerializeObject(permissionModels), Encoding.UTF8, "application/json")
+            };
+
+            await SendRequest(message).ConfigureAwait(false);
```
michaelvidal24: Should we consider changing the API to return a user object here? Seems strange that we return a no content on this one...

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -35,5 +265,17 @@ public AuthorizationClient(HttpClient client)
             var error = JsonConvert.DeserializeObject<Error>(stringResponse);
             throw new AuthorizationException(error);
         }
+
+        private async Task SendRequest(HttpRequestMessage message)
```
michaelvidal24: What happens when we have a 500 error? Are we able to Deserialize that response into an Error object?

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -15,13 +18,240 @@ public AuthorizationClient(HttpClient client)
             this.client = client;
         }
 
+        #region Users
+
+        public async Task<UserApiModel> AddUser(UserApiModel userModel)
+        {
+            var message = new HttpRequestMessage(HttpMethod.Post, new UserRouteBuilder().Route)
```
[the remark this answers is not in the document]
bsmithvt: Yep, good catch.

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -15,13 +18,240 @@ public AuthorizationClient(HttpClient client)
...
+        {
+            var message = new HttpRequestMessage(HttpMethod.Get, new UserRouteBuilder().IdentityProvider(identityProvider).SubjectId(subjectId).UserPermissionsRoute);
+            return await SendAndParseJson<List<PermissionApiModel>>(message).ConfigureAwait(false);
+        }
+
+        public async Task AddPermissionsToUser(string identityProvider, string subjectId, List<PermissionApiModel> permissionModels)
+        {
+            var message = new HttpRequestMessage(HttpMethod.Post, new UserRouteBuilder().IdentityProvider(identityProvider).SubjectId(subjectId).UserPermissionsRoute)
+            {
+                Content = new StringContent(JsonConvert.SerializeObject(permissionModels), Encoding.UTF8, "application/json")
```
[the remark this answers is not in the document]
bsmithvt: I noticed that as well. I'll see if I can refactor that easily..

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -15,13 +18,240 @@ public AuthorizationClient(HttpClient client)
...
+
         public async Task<UserApiModel> GetPermissionsForCurrentUser()
         {
-            var route = AuthorizationRoutes.GetUserPermissionUrl();
+            var route = new UserRouteBuilder().UserPermissionsRoute;
             var message = new HttpRequestMessage(HttpMethod.Get, route);
             return await SendAndParseJson<UserApiModel>(message).ConfigureAwait(false);
         }
 
+        public async Task<List<PermissionApiModel>> GetUserPermissions(string identityProvider, string subjectId)
```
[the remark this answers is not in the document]
bsmithvt: Yes - had planned to go back and add that, but I'll do that in this PR.

Fabric.Authorization.Client/AuthorizationClient.cs
michaelvidal24: Yeah - another PR definitely makes sense.

Fabric.Authorization.Client/AuthorizationClient.cs
bsmithvt: Added logic to handle this.

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -35,5 +283,26 @@ public AuthorizationClient(HttpClient client)
...
+                if (response.StatusCode < HttpStatusCode.InternalServerError)
+                {
+                    var error = JsonConvert.DeserializeObject<Error>(stringResponse);
+                    throw new AuthorizationException(error);
+                }
+
+                throw new AuthorizationException(new Error
+                {
+                    Code = response.StatusCode.ToString(),
+                    Message = "Unknown error occurred"
```
michaelvidal24: Not intending to block this PR, but we should probably either return the content of the response in the `Message`, or have the message set to "Internal Server Error".

PR #275, opened 2018-06-25
User/bsmithvt/routebuilder unit tests

bsmithvt:


Fabric.Authorization.Client/Routes/GroupRoute.cs
```diff
@@ -1,4 +1,8 @@
-﻿namespace Fabric.Authorization.Client.Routes
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Fabric.Authorization.Client.UnitTests")]
```
hc-jordenlowe: With the assembly directive.  Typically there is an AssemblyInfo.cs when you create a .net project.  In .net core there is no such thing.  Creating a class for these types of assembly level things might be better to centralize things.

hc-jordenlowe: Also, you only need one declaration per assembly.

PR #276, opened 2018-06-25
set discovery version during registration if not exists

ctediz:


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -634,7 +634,12 @@ if (!($noDiscoveryService)) {
         serviceUrl = "$authorizationServiceUrl/v1"
     }
 
-    Add-DiscoveryRegistration $discoveryServiceUrl  $credential $discoveryPostBody
+    $discoveryRegisterUrl = $discoveryServiceUrl
```
ctediz: thinking of this now, it should be enough to add '/v1' to the end instead of doing the version checks

PR #277, opened 2018-06-25
Trim ending discovery slash

ctediz:
Trim ending slash after prompting user input for discovery url so either user input or already existing discovery url has it's ending slash removed

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
@@ -499,6 +499,7 @@ if (!($noDiscoveryService)) {
         $discoveryServiceUrl = $userEnteredDiscoveryServiceUrl
     }
 
+    $discoveryServiceUrl = $discoveryServiceUrl.TrimEnd('/')
 }
 
 


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -499,6 +499,7 @@ if (!($noDiscoveryService)) {
         $discoveryServiceUrl = $userEnteredDiscoveryServiceUrl
     }
 
+    $discoveryServiceUrl = $discoveryServiceUrl.TrimEnd('/')
```
hckenmiller: This looks good, but the following code that was added previously,  if($discoveryRegisterUrl -notmatch '/v\d+/*$') {
         $discoveryRegisterUrl = "$discoveryRegisterUrl/v1"
     }
will fail if discoveryRegisterUrl was "http://localhost/DiscoveryService/v1/Services" it will return "http://localhost/DiscoveryService/v1/Services/v1"

ryanorbaker: where are we adding v1 if needed?

hckenmiller: this code is line 640 below in this script

ctediz: Passing the url as http://localhost/DiscoveryService/v1/Services, will break the fabric-install utils which should throw an error, i believe it will also break identity if it gets to the point as well since identity is also not expecting the this discovery url.  
The v1 add happens on line 638 which isn't part of this pr

hckenmiller: It is part of the last check in which I missed, since I was gone, and is based on the code I added for the bug. It is just an observation.

PR #279, opened 2018-06-26
Added option -i to switch between local code and running dockerhub image

hc-jordenlowe:
Added option -i to switch between local code and running dockerhub image

- Added a loop to parse the args for the "-i"
- Added a switch to name whether the disk comes from docker hub or local, and if it is local then build the image
-identity-setup-samples.sh needed an arg because it was using the parent script's args and that was causing issues.

### DIFF
--- a/Fabric.Authorization.API/scripts/run-client-functional-tests.sh
+++ b/Fabric.Authorization.API/scripts/run-client-functional-tests.sh
@@ -1,5 +1,27 @@
 #!/bin/bash
 
+IMAGE="0"
+
+# call script with -i to use the docker hub image
+for i in "$@"
+do
+case $i in
+    -i)
+    IMAGE="1";;
+	*)
+	echo "$0"
+   	echo "usage: $0 [-i] [--help] [-h]"
+   	echo ""
+   	echo "This script gets Identity and Authorization docker images to run tests against them.  By default the script will build the image, but if you use -i, this will pull down the image for Authorization."
+   	echo ""
+   	echo "Options:"
+   	echo "-i                  Use DockerHub Image"
+   	echo "--help, -h          Display this help page"
+   	exit 0
+    ;;
+esac
+done
+
 docker stop authz-client-functional-authorization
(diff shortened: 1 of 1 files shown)


Fabric.Authorization.API/scripts/run-client-functional-tests.sh
```diff
@@ -1,5 +1,19 @@
...
+IMAGE="0"
+
+# call script with -i to use the docker hub image
+for i in "$@"
+do
+case $i in
+    -i)
+    IMAGE="1";;
+    *)
+          # unknown option
```
michaelvidal24: Does it make sense to print out an error message here if someone attempts to use an unknown argument/option?

Fabric.Authorization.API/scripts/run-client-functional-tests.sh
```diff
@@ -23,21 +37,33 @@ echo "started identity"
...
 
+if [ "$IMAGE" -eq "1" ]; then
+	echo "Using docker image"
+	IMAGE_NAME="healthcatalyst/fabric.authorization"
+else
+	echo "Using local code"
+	IMAGE_NAME="authorization.functional.api"
+	cd ..
+	dotnet publish -o obj/Docker/publish
+	docker build -t authorization.functional.api .
```
michaelvidal24: Could we replace the image name in the `docker build` command with `$IMAGE_NAME`

ryanorbaker:
general question, wouldn't we sometimes want to use a similar setup for a an identity image?

Fabric.Authorization.API/scripts/run-client-functional-tests.sh
hc-jordenlowe: right on

Fabric.Authorization.API/scripts/run-client-functional-tests.sh
hc-jordenlowe: something like invalid args automatically print out the help arg

PR #280, opened 2018-06-27
Add Permission check methods

hc-jordenlowe:
Adding 3 methods: GetPermissionsForCurrentUser

- GetPermissionsForCurrentUser with grain and securable Items as parameters
- DoesUserHavePermissions doing a query with GetPermissionsForCurrentUser 
- DoesUserHavePermissions with UserPermissionsModel you might already have

### DIFF
--- a/Fabric.Authorization.Client/AuthorizationClient.cs
+++ b/Fabric.Authorization.Client/AuthorizationClient.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
@@ -14,11 +15,11 @@ namespace Fabric.Authorization.Client
     public class AuthorizationClient
     {
         
-        private readonly HttpClient client;
+        private readonly HttpClient _client;
 
         public AuthorizationClient(HttpClient client)
         {
-            this.client = client;
+            this._client = client;
         }
 
         #region Users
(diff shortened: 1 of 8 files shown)


Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -34,16 +35,41 @@ public AuthorizationClient(HttpClient client)
...
             var message = new HttpRequestMessage(HttpMethod.Get, route)
                 .AddBearerToken(accessToken)
                 .AddAcceptHeader(ClientConstants.ApplicationJson);
 
             return await SendAndParseJson<UserPermissionsApiModel>(message).ConfigureAwait(false);
         }
-        
+
+        public async Task<bool> DoesUserHavePermission(string accessToken, string permission)
+        {
```
bsmithvt: General comment and I'm not sure what others' thoughts are, but should we just pass through whatever the user passes in and let the API do the validation of parameters? So essentially remove the `CheckIf*` calls?

michaelvidal24: I think checking for null/empty strings is fine - it will help to keep clients from having to go to the server to find out that they didn't pass an access token, which reduces the time it takes to get an error response and lowers the load on the service itself. I wouldn't be any deeper validation on the client though, as we want to keep that logic with the API itself.

Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -275,7 +300,7 @@ public AuthorizationClient(HttpClient client)
 
         #endregion
 
-        private void CheckIfNull(string value, string name)
+        private void CheckIfStringNullOrEmpty(string value, string name)
```
bsmithvt: Should an exception be thrown below?

Fabric.Authorization.Client/AuthorizationClient.cs
bsmithvt: That works.

hc-jordenlowe: Agreed

Fabric.Authorization.Client/AuthorizationClient.cs
hc-jordenlowe: there is an AuthorizationException being thrown.

PR #281, opened 2018-06-28
Changed .net 2.0 to .net 1.1

hc-jordenlowe:


doc/arch/ADR-0001 CreatedAuthorizationClient.md
```diff
@@ -0,0 +1,30 @@
+# Title
+
+Created Authorization Client
```
ryanorbaker: should we mention c# here?  and maybe talk about why c# vs powershell?

hc-jordenlowe: I dont think the language decision is something that needs to be discussed.  Of course if there is some emphasis we want to stress then I can put that down.  Im just not following what we want to talk about C# for.

What would we need to talk about the SDK being a library instead of a powershell SDK?  If I remember, its because we will be integrating custom applications and this will help with that adpotion, internally and externally.  Is there any more than can be added to this?

PR #282, opened 2018-07-09
Added nuget target to include project references properly

hc-jordenlowe:
Added code that will package project references using a build target.  The target works by looking for PrivateAssets="all", which indicates it is part of this package (Not something that is public like in a nuget repo.).  If this is not included, then the nuget pack will assume the reference is in nuget repo and it will not pull in those resources.

### DIFF
--- a/Fabric.Authorization.Client/Fabric.Authorization.Client.csproj
+++ b/Fabric.Authorization.Client/Fabric.Authorization.Client.csproj
@@ -2,6 +2,21 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.6</TargetFramework>
+    <Authors>Fabric.Authorization.Client</Authors>
+    <Company>Health Catalyst</Company>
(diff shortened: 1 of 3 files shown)


Fabric.Authorization.Client/Fabric.Authorization.Client.csproj
```diff
@@ -2,14 +2,43 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.6</TargetFramework>
+    <Authors>Fabric.Authorization.Client</Authors>
+    <Company>Health Catalyst</Company>
+    <Version>1.0.0.0-local</Version>
```
ryanorbaker: does the normal build process also update this minor/sub versions?

hc-jordenlowe: Yes, this will be changed with some command line magic.  I copied these values from our other projects, but they will be transformed

PR #285, opened 2018-07-12
Added versioning

hc-jordenlowe:


johnrulon:
I thought the build itself could update the assembly version/info? I don't understand why we need to make code change, but if others do then ignore my comment and move ahead :)

johnrulon:
@michaelvidal24 ah...makes sense. Nice!

PR #291, opened 2018-07-25
User/bsmithvt/access control script

bsmithvt:


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -758,4 +757,25 @@ Write-Host "There are $($edwAdminUsers.Count) users with this role"
...
 
+#Register Fabric.Authorization.AccessControl client
+$body = @'
+{
+    "clientId":"fabric-access-control", 
+    "clientName":"Fabric Authorization Access Control Client", 
+    "requireConsent":"false", 
+    "allowedGrantTypes": ["implicit"], 
+    "allowedScopes": ["openid", "profile", "fabric.profile", "fabric/authorization.read", "fabric/authorization.write", "fabric/idprovider.searchusers", "fabric/authorization.dos.write"],
+	"allowedCorsOrigins": ["$authorizationServiceUrl/v1"],
```
michaelvidal24: The allowed cors origins should just be the URL of the root web server.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -758,4 +757,25 @@ Write-Host "There are $($edwAdminUsers.Count) users with this role"
...
+	"redirectUris": ["$authorizationServiceUrl/v1/oidc-callback.html", "$authorizationServiceUrl/v1/silent.html"],
```
michaelvidal24: Do we need the v1's here? I don't think users should need to put in a /v1 to get to the UI, they should just be able to enter something like https://host.domain.local/Authorization. The UI should know if its using a v1 or v2 endpoints of the API.

Fabric.Authorization.API/Modules/RootModule.cs
```diff
@@ -11,7 +12,7 @@ public RootModule(): base("/v1")
 
         private dynamic Redirect()
         {
-            return Response.AsRedirect("/client/index.html");
+            return Response.AsRedirect($"/{AccessControl.Path}/index.html");
```
michaelvidal24: Will this allow us to hit the app using https://host.domain.local/Authorization, or do we have to enter:
https://host.domain.local/Authorization/accesscontrol?

bsmithvt: Yes, when I run locally and go to `localhost:5004`, I get redirected to the landing page for the access control UI.

PR #293, opened 2018-07-25
added initial screen for left hand navigation 

hc-jordenlowe:
I have created a left handed nav with sub items.  

Tried the cashmere list, but it felt like it was restrictive on multiline items.  You had to explicited label each line.  If this gets recursive, this may be a big issue.  Of course if there is some advice, let me know.  Link to the list:  https://cashmere.healthcatalyst.net/components/list

Also, this is just a shell.  I dont have the service to call the API yet.  I just wanted to make sure the general look was acceptable before proceeding.

### DIFF
--- a/Fabric.Authorization.API/Modules/GrainsModule.cs
+++ b/Fabric.Authorization.API/Modules/GrainsModule.cs
@@ -1,0 +1,37 @@
+﻿using Fabric.Authorization.API.Services;
+using Fabric.Authorization.API.Models;
+using Fabric.Authorization.Domain.Models;
+using Fabric.Authorization.Domain.Services;
+using Nancy.Security;
+using Serilog;
+using System;
+using System.Threading.Tasks;
+using Fabric.Authorization.Domain.Validators;
+
+namespace Fabric.Authorization.API.Modules
+{
+    public class GrainsModule : FabricModule<Grain>
+    {
+        private readonly GrainService _grainService;
+
+        public GrainsModule(GrainService grainService,
+            GrainValidator validator,
+            AccessService accessService,
+            ILogger logger) : base("/v1/grains", logger, validator, accessService)
+        {
+            _grainService = grainService ??
+                           throw new ArgumentNullException(nameof(grainService));
+
+            Get("/",
+                async _ => await GetGrain().ConfigureAwait(false),
+                null,
+                "GetGrain");
+        }
+
+        private async Task<dynamic> GetGrain()
+        {
+            CheckReadAccess();
+            return (await _grainService.GetAllGrains()).ToGrainApiModels();
+        }
+    }
+}
--- a/Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGrainStore.cs
+++ b/Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGrainStore.cs
@@ -38,6 +38,28 @@ namespace Fabric.Authorization.Persistence.SqlServer.Stores
             return grain.ToModel();
         }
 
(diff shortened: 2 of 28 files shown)


Fabric.Authorization.API/Modules/GrainsModule.cs
```diff
@@ -0,0 +1,38 @@
...
+            Get("/",
+                async _ => await GetGrain().ConfigureAwait(false),
+                null,
+                "GetGrain");
+        }
+
+        private async Task<dynamic> GetGrain()
+        {
+            CheckReadAccess();
+            this.RequiresClaims(AuthorizationReadClaim);
```
michaelvidal24: The CheckReadAccess() function is a helper function that does what this.RequireClaims(AuthorizationReadClaim) does, we only need the CheckReadAccess() statement.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGrainStore.cs
```diff
@@ -38,6 +38,28 @@ public class SqlServerGrainStore : SqlServerBaseStore, IGrainStore
...
+        {
+            var grains = AuthorizationDbContext.Grains
+                .Include(g => g.SecurableItems)
+                .Where(g => !g.IsDeleted);
+
+            foreach (var grain in grains)
+            {
+                foreach (var securableItem in grain.SecurableItems)
+                {
+                    LoadChildrenRecursive(securableItem);
```
michaelvidal24: We may need to update this function to filter our deleted securables.

Fabric.Authorization.API/Modules/GrainsModule.cs
hc-jordenlowe: gotcha.  Good to know

hc-jordenlowe:
I have rebased.  I will add the feature flag.  Then the last thing will be to add tests and this section should be done.  Next bits on the task list:

1. Wire up the left pane with the right, this includes the highlighting
2. add expand/contract buttons for the lists of grains/securable items
3. add deep linking
4. see if we can migrate to models project

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGrainStore.cs
hc-jordenlowe: This function is used within our securableItems code as well.  I would need to create something that is a separate function.

bsmithvt: Perhaps that's a bug in the existing code? Is there ever a case where we need to return deleted ones?

michaelvidal24: That is what I was thinking, we can just look to see what is using this function and determine if the current usages need deleted items or not.

bsmithvt: Tracing up the call stack, it's used in `GetRequiredWriteClaims(GrainService grainService, string grain)` in `FabricModule` and when adding a client. I'm thinking they *should* be filtered out for these scenarios.

michaelvidal24: Agreed.

hc-jordenlowe: kk, i will filter deleted stuff.  I went back to look for the other usage, and i cant find it.  Thought Securable items was working the same way. 

Ah well, added this code and if i run into this for securable items, i will change it.

PR #295, opened 2018-07-26
User/bsmithvt/access control config

bsmithvt:


Fabric.Authorization.AccessControl/src/app/services/global/auth.service.ts
```diff
@@ -16,13 +16,13 @@ export class AuthService {
...
-    this.clientId = 'fabric-accesscontrol';
+    this.clientId = 'fabric-access-control';
     this.authority = environment.fabricIdentityApiUri;
 
     const clientSettings: any = {
       authority: this.authority,
       client_id: this.clientId,
-      redirect_uri: `${environment.applicationEndpoint}/oidc-callback.html`,
+      redirect_uri: `${environment.applicationEndpoint}/client/oidc-callback.html`,
       post_logout_redirect_uri: environment.applicationEndpoint,
```
michaelvidal24: I think this line needs to be updated to reflect the new logged-out redirect uri.

PR #297, opened 2018-07-27
Corrected forward slash issue.

hc-jordenlowe:
This corrects the issue where forward slashes cause processing incorrect urls.

1. base url needs to have a forward slash
2. relative url paths should not have a forward slash at the beginning

I have updated the unit tests to reflect both scenarios.

### DIFF
--- a/Catalyst.Fabric.Authorization.Client/Extensions/AuthorizationExtensions.cs
+++ b/Catalyst.Fabric.Authorization.Client/Extensions/AuthorizationExtensions.cs
@@ -1,11 +1,13 @@
-﻿using System.Net.Http;
+﻿using System;
+using System.Linq;
+using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text;
 using Newtonsoft.Json;
 
 namespace Catalyst.Fabric.Authorization.Client.Extensions
 {
-    internal static class HttpRequestMessageExtensions
+    internal static class AuthorizationExtensions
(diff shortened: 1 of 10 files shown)


Catalyst.Fabric.Authorization.Client/Extensions/AuthorizationExtensions.cs
```diff
@@ -24,5 +26,15 @@ public static HttpRequestMessage AddContent<T>(this HttpRequestMessage httpReque
...
             return httpRequestMessage;
         }
+
+        public static HttpClient FormatBaseUrl(this HttpClient client)
+        {
+            // so this will trim the forward slash whether its there or not.  it will 
+            // also fix any additional forward slashes.  Once it has trimmed the end,
+            // it will append the forward slash needed for making accurate calls.
+            // see also: https://stackoverflow.com/questions/23438416/why-is-httpclient-baseaddress-not-working#23438417
+            client.BaseAddress = new Uri($"{client.BaseAddress.OriginalString.TrimEnd('/')}/");
```
ryanorbaker: we ended up doing the same thing in the Identity CLI

hc-jordenlowe: its simple code to just make sure the right forward slash is there.  Kinda lame but i dont think it matters too much to make it simple.

PR #298, opened 2018-07-30
Fixed up the handling of the logout route when the app is hosted via Nancy.

michaelvidal24:


Fabric.Authorization.API/Modules/RootModule.cs
```diff
@@ -8,6 +8,7 @@ public class RootModule : NancyModule
         public RootModule(): base("/v1")
         {
             Get("/", _ => Redirect(), null, "Redirect");
```
bsmithvt: Can we use the `AccessControl.Path` constant here?

michaelvidal24: Good call, I'll update that.

PR #300, opened 2018-07-31
User domain doesn't always match computer domain

nathan-given-health-catalyst:
suggesting a fix for the issue that the user domain doesn't always match the computer domain


nathan-given-health-catalyst:
Yeah... my copy/paste got messed up.  It needs to have a space

    $currentDomain = (Get-WmiObject Win32_ComputerSystem).Domain

PR #301, opened 2018-08-03
Added retry and circuitbreaker logic

hc-jordenlowe:
Needs some testing and attention to the exception scenarios that we need to capture, but this is a foundation for introducing Polly (retry and circuit breaker logic) into our API.

I added an optional cancellation token as well to match what HttpClient does for it's web requests.

Basically, if you are a 500, a cancelled task or timeout, then we should try to retry the request with an exponential backoff (i.e. 1, 2, 4, 8, etc).

Now, if you encounter so many failures within a minute (like 200) then we should stop processing a request and throw a circuit breaker exception.  This is so we dont DDOS our service.  After a cool down, then it will test to make sure it can access the resource again.

Anyways, this is a start.  perhaps something to make a story out of it.

### DIFF
--- a/Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
+++ b/Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
@@ -3,38 +3,90 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Catalyst.Fabric.Authorization.Client.Extensions;
 using Catalyst.Fabric.Authorization.Client.Routes;
 using Catalyst.Fabric.Authorization.Models;
 using Catalyst.Fabric.Authorization.Models.Requests;
 using Newtonsoft.Json;
+using Polly;
 
 namespace Catalyst.Fabric.Authorization.Client
 {
     public class AuthorizationClient
     {
-        
+        private Policy RetryPolicy;
+        private Policy CircuitBreakerPolicy;
+
+        private int MaxRetryAttempts;
+
+        private int TimeSpanInMinutes;
+
+        private int NumberOfErrorsBeforeCircuitBreak;
+
         private readonly HttpClient _client;
 
-        public AuthorizationClient(HttpClient client)
+        public AuthorizationClient(HttpClient client, bool useCircuitBreaker = false, bool useRetry = false)
         {
             this._client = client.FormatBaseUrl();
+            
+            if(useCircuitBreaker)
+            {
+                this.InitializeCircuitBreaker();
+            }
+
+            if(useRetry)
+            {
+                this.InitializeRetry();
+            }
         }
-        
+
+        public AuthorizationClient(HttpClient client, bool useCircuitBreaker = false, bool useRetry = false, int maxRetryAttempts = 5, int timeSpanInMinutes = 1, int numberOfErrorsBeforeCircuitBreak = 200)
+            : this(client, useCircuitBreaker, useRetry)
+        {
+            this.MaxRetryAttempts = maxRetryAttempts;
+            this.TimeSpanInMinutes = timeSpanInMinutes;
+            this.NumberOfErrorsBeforeCircuitBreak = numberOfErrorsBeforeCircuitBreak;
+        }
+
+        private void InitializeCircuitBreaker()
+        {
+            // if there is any Authorization Exception,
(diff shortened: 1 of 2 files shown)


Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -3,38 +3,90 @@
...
 using Newtonsoft.Json;
+using Polly;
 
 namespace Catalyst.Fabric.Authorization.Client
 {
     public class AuthorizationClient
     {
-        
+        private Policy RetryPolicy;
+        private Policy CircuitBreakerPolicy;
```
michaelvidal24: Since this is a local policy, then the circuit will only break when this instance of the policy hits its limit, is that what we want? To be most effective, it would require the client of this library only instantiating one AuthorizationClient.

Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -3,38 +3,90 @@
...
+        private int MaxRetryAttempts;
+
+        private int TimeSpanInMinutes;
+
+        private int NumberOfErrorsBeforeCircuitBreak;
+
         private readonly HttpClient _client;
 
-        public AuthorizationClient(HttpClient client)
+        public AuthorizationClient(HttpClient client, bool useCircuitBreaker = false, bool useRetry = false)
```
michaelvidal24: should we consolidate the constructor here and simply specify a boolean like `useConnectionResiliancy`? I would imaging if someone wants resiliancy, they would want both the circuit breaker and the retry policies.

Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -3,38 +3,90 @@
...
+            }
+
+            if(useRetry)
+            {
+                this.InitializeRetry();
+            }
         }
-        
+
+        public AuthorizationClient(HttpClient client, bool useCircuitBreaker = false, bool useRetry = false, int maxRetryAttempts = 5, int timeSpanInMinutes = 1, int numberOfErrorsBeforeCircuitBreak = 200)
```
michaelvidal24: Could we create a struct or a class that encapsulates the config for retryattempts, errors and time span?

Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -3,38 +3,90 @@
...
+
+        private void InitializeCircuitBreaker()
+        {
+            // if there is any Authorization Exception,
+            // AggregrateException or OptionationCanceledException,
+            // and there are 200 in one minute, it will trigger the
+            // circuit breaker;
+            this.CircuitBreakerPolicy = Policy.Handle<OperationCanceledException>()
+                .Or<AuthorizationException>()
+                .Or<OperationCanceledException>()
```
michaelvidal24: Is this repeated from line 59?

Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -3,38 +3,90 @@
...
+
+        private void InitializeRetry()
+        {
+            // if the exception is a 500 server error of any kind,
+            // do a retry just to make sure.
+            // Everything else, just fail.
+            // if there are too many 500 errors, then circuit break.
+            this.RetryPolicy = Policy.Handle<OperationCanceledException>()
+                .Or<AuthorizationException>(auth => auth.Details.Code.First() == '5' && auth.Details.Code.Length == 3)
+                .Or<OperationCanceledException>()
```
michaelvidal24: Is this repeated from line 71?

Catalyst.Fabric.Authorization.Client/AuthorizationClient.cs
```diff
@@ -340,9 +392,33 @@ private void CheckIfNull(object value, string name)
...
         }
 
-        private async Task<T> SendAndParseJson<T>(HttpRequestMessage message)
+        private async Task<T> SendAndParseJson<T>(HttpRequestMessage message, CancellationToken? token)
+        {
+            // This function wraps all the retry/circuit breaker logic
+            // for the sending of a request.
+            
+            // retry first a few times.
+            var RetryTask = RetryPolicy.ExecuteAsync<T>(
```
michaelvidal24: Should we use the PolicyWrap policy so the circuit breaker can be nested inside the retry?

PR #302, opened 2018-08-03
Added treeview and Angular 6

hc-jordenlowe:
Two things, Angular 6 and Treeview:

Per suggestion from Data Governance, a great suggestion was to use Angular Material to render a tree view.  Works great except that I also need angular 6.  I have upgraded the project but there might be more regression to do if we want to go this route.


Fabric.Authorization.AccessControl/src/environments/environment.default.ts
```diff
@@ -3,8 +3,8 @@ export const defaultEnvironment = {
   production: false,
   fabricAuthApiUri: 'http://localhost:5004',
   fabricAuthApiVersionSegment: 'v1',
-  fabricIdentityApiUri: 'http://localhost/identity',
-  fabricExternalIdPSearchApiUri: 'http://localhost/IdentityProviderSearchService',
+  fabricIdentityApiUri: 'https://HC2282.hqcatalyst.local/identity',
```
michaelvidal24: remember not to check these changes in.

hc-jordenlowe: your right, let me remove this now.

Fabric.Authorization.AccessControl/karma.conf.js
```diff
@@ -6,24 +6,22 @@ process.env.CHROME_BIN = require("puppeteer").executablePath();
 module.exports = function (config) {
   config.set({
     basePath: '',
-    frameworks: ['jasmine', '@angular/cli'],
+    frameworks: ['jasmine', '@angular-devkit/build-angular'],
```
ryanorbaker: is assuming CLI built into build-angular?  assuming that's part of Angular 6?

hc-jordenlowe: right, but let me double check that is the case.

Fabric.Authorization.API/Startup.cs
```diff
@@ -30,12 +30,7 @@ public class Startup
             "",
             "/",
             $"/{AccessControl.Path}",
-            $"/{AccessControl.Path}/",
-            $"/{AccessControl.Path}/index.html",
-            $"/{AccessControl.Path}/oidc-callback.html",
-            $"/{AccessControl.Path}/silent.html",
-            $"/{AccessControl.Path}/logged-out",
-            $"/{AccessControl.Path}/access-control"
+            $"/{AccessControl.Path}/*"
```
michaelvidal24: This is so much cleaner!

Fabric.Authorization.AccessControl/src/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <title>Fabric.Authorization.AccessControl</title>
-  <base href="/">
+  <base href="CLIENT_ROOT">
```
michaelvidal24: When we run ng build or npm run build, will this get replaced? If it does, then it will break the replacement logic in the middleware component.

Fabric.Authorization.AccessControl/karma.conf.js
hc-jordenlowe: After looking up the googles, looks like this is a result of the angular v6 migration.  It is suppose to be there.

Fabric.Authorization.AccessControl/src/index.html
michaelvidal24: We could just make the token in the middleware `<base href="/">`

hc-jordenlowe: So this works fine.  Running an "ng build" will not replace it.  When our midware runs, then that will replace it with the base url of the running site.  

Now, if we run "ng build --base-href=..."  This is the case where it will replace that value, however I removed this from our setup scripts.  We will be fine if we leave everything as "ng build" or "npm run build"

michaelvidal24: Has it also been removed from our build process?

hc-jordenlowe: i did not check that.  I can do so

hc-jordenlowe: Looks like we just run npm build in the build phase.

In the release phase, there are no other things running, besides the npm packaging for Access Control.

Looks like its good on that front.  I will leave it open until after standup.  Meanwhile I will branch off it and do some unit tests.

PR #304, opened 2018-08-06
Fixed the redirect to the access-control UI so it works regardless of hosting configuration.

michaelvidal24:


Fabric.Authorization.API/Modules/RootModule.cs
```diff
@@ -1,19 +1,24 @@
...
+        private readonly IAppConfiguration _appConfiguration;
+        public RootModule(IAppConfiguration appConfiguration): base("/v1")
         {
+            _appConfiguration = appConfiguration;
             Get("/", _ => Redirect(), null, "Redirect");
             Get($"/{AccessControl.Path}", _ => Redirect(), null, "Redirect");
         }
 
         private dynamic Redirect()
         {
```
bsmithvt: Can we use `IHostingEnvironment.WebRootPath` here to avoid the need for a new configuration variable (similar to what I did in https://github.com/HealthCatalyst/Fabric.Authorization/blob/27af629a9b98963380a4b1e7e6205585c5fffc55/Fabric.Authorization.API/Infrastructure/Middleware/AngularMiddleware.cs)?

michaelvidal24: I'll test that out and see.

michaelvidal24: WebRootPath doesn't provide a URL, it provides a file path to the web content, in this case we are redirecting to a URL, so we can't take advantage of the file path.

bsmithvt: @michaelvidal24, that's right. Sorry about that.

PR #306, opened 2018-08-09
Ensure discovery service call uses credentials. Serve index.html from…

michaelvidal24:
… memory rather than file.

### DIFF
--- a/Fabric.Authorization.API/Infrastructure/Middleware/AngularMiddleware.cs
+++ b/Fabric.Authorization.API/Infrastructure/Middleware/AngularMiddleware.cs
@@ -38,18 +38,40 @@ namespace Fabric.Authorization.API.Infrastructure.Middleware
                 var discoveryServiceSettings = _appConfiguration.AccessControlSettings.DiscoveryServiceSettings;
                 _indexContent = _indexContent.Replace(discoveryServiceSettings.Token,
                     discoveryServiceSettings.Value);
-
-                File.WriteAllText(fullPath, _indexContent);
             }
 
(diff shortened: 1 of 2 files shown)


Fabric.Authorization.API/Infrastructure/Middleware/AngularMiddleware.cs
```diff
@@ -38,18 +38,40 @@ public async Task Invoke(HttpContext context)
                 var discoveryServiceSettings = _appConfiguration.AccessControlSettings.DiscoveryServiceSettings;
                 _indexContent = _indexContent.Replace(discoveryServiceSettings.Token,
                     discoveryServiceSettings.Value);
-
-                File.WriteAllText(fullPath, _indexContent);
             }
```
bsmithvt: Should this be case-insensitive comparison?

Fabric.Authorization.API/Infrastructure/Middleware/AngularMiddleware.cs
```diff
@@ -38,18 +38,40 @@ public async Task Invoke(HttpContext context)
...
+            }
+        }
+
+        private async Task WriteIndexResponse(HttpContext context)
+        {
+            using (var memoryStream = new MemoryStream(Encoding.ASCII.GetBytes(_indexContent)))
+            {
+                var originalResponse = context.Response.Body;
+                try
+                {
```
bsmithvt: Does the response body byte size need to be adjusted, or does it automatically get set correctly? I remember reading something about that online when I was working on this.

PR #307, opened 2018-08-13
User/bsmithvt/custom ad groups

bsmithvt:


bsmithvt:
@michaelvidal24, good call, but we would need to enforce it only for cases where `IsDeleted=0`. Did we do this somewhere already?

michaelvidal24:
Yes, we have an example of that in Fabric.Identity: https://github.com/HealthCatalyst/Fabric.Identity/blob/master/Fabric.Identity.SqlServer/Tables/Clients.sql.

bsmithvt:
One thing I still need to do is handle multiple levels of group nesting. Currently only 1 level of nesting is supported.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
```diff
@@ -341,6 +356,65 @@ public async Task<Group> DeleteUserFromGroup(Group group, User user)
...
+                childToRemoveEntity.IsDeleted = true;
+            }
+
+            await AuthorizationDbContext.SaveChangesAsync();
+            await EventService.RaiseEventAsync(new EntityAuditEvent<Group>(EventTypes.ChildEntityDeletedEvent, group.Id.ToString(), group));
+
+            return group;
+        }
+
+        public Task<IEnumerable<Group>> Get(IEnumerable<string> childGroupNames)
```
michaelvidal24: Should we name this parameter something other than childGroupNames? Seems like this function is good at getting a list of groups based on a list of group names and is not really specific to childGroups...

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
```diff
@@ -126,6 +140,7 @@ public async Task<IEnumerable<Group>> GetAll()
                 .ThenInclude(gr => gr.Role)
                 .Include(g => g.GroupUsers)
                 .ThenInclude(gu => gu.User)
+                .Include(g => g.ChildGroups)
                 .Where(g => !g.IsDeleted)
                 .ToArrayAsync();
```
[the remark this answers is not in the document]
bsmithvt: Good call. Will change it.

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
bsmithvt: Good call. Will change it.

PR #313, opened 2018-08-20
User/bsmithvt/sdk updates

bsmithvt:


Fabric.Authorization.IntegrationTests/Modules/ChildGroupsTests.cs
```diff
@@ -166,7 +166,7 @@ public async Task AddChildGroup_ValidRequest_SuccessAsync()
                 });
             });
 
-            Assert.Equal(HttpStatusCode.Created, postResponse.StatusCode);
+            Assert.Equal(HttpStatusCode.OK, postResponse.StatusCode);
```
hc-jordenlowe: For posting, wouldnt you want Created?

bsmithvt: In this case, we're creating an association between 2 resources (groups), so conceptually we're updating 2 existing groups. I return a 200 with the modified parent group in the response. This is consistent with the other auth endpoints that create "association" records.

hc-jordenlowe: cool, good to know.

PR #317, opened 2018-08-22
Uri class will not append the url correctly. 

hc-jordenlowe:
Changed code to get the right client url.

Any suggestions on how to test this.  I will try to get something going by the morning.


hc-jordenlowe:
Writing a unit test would have been difficult.  Instead I tested these scenarios:

1. http://localhost
2. http://localhost/
3. https://hc2282.hqcatalyst.local/Authorization
4. https://hc2282.hqcatalyst.local/Authorization/extra/

With each item, the client was appended on as expected.  This should be good enough for a thumbs up.

PR #318, opened 2018-08-23
Dont prompt for app pool user if app pool already exists

michaelvidal24:


bsmithvt:
Can lines 783-817 be put in the shared PS module and used by Identity and Auth scripts?

michaelvidal24:
@bsmithvt - I'm planning on addressing that when I add in the ability to do quiet mode.

PR #319, opened 2018-08-23
User/jasonsoto/rolemanager

ctediz:
Migration of roles to EDWAdmin


Fabric.Authorization.Domain/Services/RoleManager.cs
```diff
@@ -0,0 +1,100 @@
+﻿namespace Fabric.Authorization.Domain.Services
```
bsmithvt: Since we have a `RoleService`, I would suggest naming this something like `EdwAdminSyncService` or something more specific so that people don't get confused on the intent of this class.

Fabric.Authorization.Persistence.SqlServer/Stores/EDW/ISecurityContext.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Catalyst.Security.Services
```
bsmithvt: Namespace doesn't match folder structure?

Fabric.Authorization.Persistence.SqlServer/Stores/EDW/SecurityContext.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Catalyst.Security.Services;
+using Fabric.Authorization.Persistence.SqlServer.Configuration;
+using Microsoft.EntityFrameworkCore;
+
+namespace Fabric.Authorization.API.Models.EDW
```
bsmithvt: Namespace doesn't match folder structure?

Fabric.Authorization.Persistence.SqlServer/Stores/EDW/ISecurityContext.cs
bsmithvt: What's the reasoning behind the name `ISecurityContext` vs. `IMetadataDatabaseContext` or something like that?

bsmithvt:
One other general comment - I would suggest adding some integration tests around this.

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -228,13 +243,21 @@ public async Task<Group> AddChildGroups(Group group, IEnumerable<string> childGr
                 throw new AlreadyExistsException<Group>($"The following groups are already children of group {group.Name}: {string.Join(", ", existingAssociations.Select(g => g.Name))}");
             }
 
-            return await _groupStore.AddChildGroups(group, childGroups);
+            var persistedGroup = await _groupStore.AddChildGroups(group, childGroups);
+            await this._roleManager.RefreshDosAdminRolesAsync(group.Children.SelectMany(childGroup => childGroup.Users));
```
bsmithvt: Do we need this in the child groups logic? This is just removing the association between a parent and child group.

ctediz: Ahh I see, was under the impression that the child group could be a new group, but looking at it now it takes in names rather than an actuall group

ctediz: How do the roles relate to users here, for example, if a user is in the parent group and a child group is added, do the users from the parent group inherit the child group roles as well?

Fabric.Authorization.API/DependencyInjection/InMemoryConfigurator.cs
```diff
@@ -14,5 +15,11 @@ protected override TinyIoCContainer.RegisterOptions RegisterDatabaseContext(Tiny
         {
             return container.Register<IAuthorizationDbContext, InMemoryAuthorizationDbContext>();
         }
+
+        protected override TinyIoCContainer.RegisterOptions RegisterEDWDatabaseContext(TinyIoCContainer container)
+        {
+            throw new NotImplementedException("There is no In memory EDW Database context.");
```
michaelvidal24: If we don't provide an InMemoryEdwContext, then our in memory integration tests will not work properly.

Fabric.Authorization.API/DependencyInjection/BaseSqlServerConfigurator.cs
```diff
@@ -20,6 +20,7 @@ public void ConfigureApplicationInstances(TinyIoCContainer container)
         {
             container.Register<IDbBootstrapper, SqlServerDbBootstrapper>().AsMultiInstance();
             RegisterDatabaseContext(container).AsMultiInstance();
+            RegisterEDWDatabaseContext(container).AsMultiInstance();
```
michaelvidal24: Do we need an EdwDatabaseContext in the application container? We should only need this if we want to do work with the context outside the scope of a request, for instance at startup.

Fabric.Authorization.Domain/Services/EDWAdminRoleSyncService.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            else
+            {
+                _edwStore.RemoveIdentitiesFromRole(new[] { user.SubjectId }, EDWConstants.EDWAdmin);
+            }
+        }
+
+        public async Task RefreshDosAdminRolesAsync(IEnumerable<User> users)
+        {
+            foreach(User user in users)
```
michaelvidal24: Looks like we are hitting the database on each loop - is there a more efficient way to do this?

Fabric.Authorization.Domain/Services/EDWAdminRoleSyncService.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                return true;
+            }
+
+            // next, get all the groups for a user
+            // if a group is a special Super Admin Group then return true
+            // if not, then look on each group to see if there is a "super admin role"
+            foreach (var customGroup in user.Groups)
+            {
+                var group = await _groupService.GetGroup(customGroup.Name).ConfigureAwait(false);
```
michaelvidal24: We now have a GetGroups function, it might be more efficient to use that in this case so we can get all the groups we need in one shot instead of hitting the database in a loop.

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -32,22 +32,28 @@ public class GroupService
...
             IUserStore userStore,
-            RoleService roleService)
+            RoleService roleService,
+            EDWAdminRoleSyncService roleManager)
         {
             _roleStore = roleStore;
             _userStore = userStore;
             _groupStore = groupStore ?? throw new ArgumentNullException(nameof(groupStore));          
             _roleService = roleService ?? throw new ArgumentNullException(nameof(roleService));
+            _roleManager = roleManager ?? throw new ArgumentNullException(nameof(roleManager));
```
michaelvidal24: I know we renamed the class, should we also rename the private variable for clarity?

Fabric.Authorization.Persistence.SqlServer/Services/InMemorySecurityContext.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class InMemorySecurityContext : SecurityContext
+    {
+        public InMemorySecurityContext()
+        {
+        }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            //optionsBuilder.UseInMemoryDatabase(databaseName: ConnectionStrings.AuthorizationDatabase);
```
michaelvidal24: Can we make a valid in memory context here so we can test this appropriately in our integration tests?

Fabric.Authorization.Persistence.SqlServer/Stores/EDW/ISecurityContext.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using Fabric.Authorization.Domain.Models.EDW;
+using Microsoft.EntityFrameworkCore;
+
+namespace Fabric.Authorization.Persistence.SqlServer.Stores.EDW
+{
+    public interface ISecurityContext : IDisposable
```
michaelvidal24: I don't think we need to specify IDisposable here...

Fabric.Authorization.Persistence.SqlServer/Stores/EDW/SecurityContext.cs
```diff
@@ -0,0 +1,47 @@
...
+            //    .HasOne(identityRole => identityRole.EDWRole)
+            //    .WithMany(role => role.EDWIdentityRoles)
+            //    .HasForeignKey(identityRole => identityRole.RoleId);
+            //modelBuilder.Entity<EDWIdentityRole>()
+            //    .HasOne(identityRole => identityRole.EDWIdentity)
+            //    .WithMany(role => role.EDWIdentityRoles)
+            //    .HasForeignKey(identityRole => identityRole.RoleId);
+        }
+
+        public ISecurityContext CreateContext()
```
michaelvidal24: If we provide the appropriate constructor that takes in a ConnectionStrings object we should be able to get rid of this method and simply allow the DI system to create the context for us, then we also don't have to worry about disposing it, the lifetime will get managed by the DI system.

Fabric.Authorization.Persistence.SqlServer/Stores/EDWStore.cs
```diff
@@ -0,0 +1,95 @@
...
+
+
+        private IQueryable<EDWRole> GetRolesByName(string roleName)
+        {            
+            return this.securityContext.EDWRoles.Where(role => role.Name == roleName);
+        }
+
+        private void AddIdentitiesToRole(string[] identityNames, int roleId)
+        {
+            using (var context = this.securityContext.CreateContext())
```
michaelvidal24: We should just be able to use the context, and not have to create a new one for each call and worry about disposing it.

Fabric.Authorization.Persistence.SqlServer/Stores/EDWStore.cs
```diff
@@ -0,0 +1,95 @@
...
+                    }
+                }
+
+                context.SaveChanges();
+
+                var roleToAdd = context.EDWRoles.FirstOrDefault(role => role.Id == roleId);
+
+                if (roleToAdd == null)
+                {
+                    throw new ArgumentException("Role not found.");
```
michaelvidal24: I think a more informative error message here would be helpful. Since the app is using two context and each has a concept of role, we should at least let people know which role we are looking from which context.

michaelvidal24: Should we also consider some sort of partial success, or perhaps wrap the whole thing in a transaction? What it looks like can happen now is, the auth database gets updated, but this update fails, and now we are inconsistent...

Fabric.Authorization.Persistence.SqlServer/Stores/EDWStore.cs
```diff
@@ -0,0 +1,95 @@
...
+        private void RemoveIdentitiesFromRole(string[] identityNames, int roleId)
+        {
+            using (var context = this.securityContext.CreateContext())
+            {
+                var targetIdentities = securityContext.EDWIdentities.Include(identity => identity.EDWRoles)
+                .Where(identity => identityNames.Contains(identity.Name));
+
+                foreach (var identity in targetIdentities)
+                {
+                    var roleToRemove = identity.EDWRoles.FirstOrDefault(role => role.Id == roleId);
```
michaelvidal24: If we move this out of the loop, we would cut down on db calls.

Fabric.Authorization.Domain/Services/GroupService.cs
bsmithvt: Yes, users will inherit the permissions of the roles tied to the child groups. Users can only be added to `custom` groups, and only `custom` groups can be parents,, while only `directory` (i.e., AD) groups an be children. So perhaps this (and the sync logic in `RemoveChildGroups`) does need to stay now that I think about it. However, I don't think the logic in the query is correct:

1) `ChildGroup` navigation collection does not include `Users`, but this is irrelevant since AD groups cannot have users.
2) I think the `IsUserASuperAdmin` method needs to be changed to check the group's children's roles for the 2 roles we want to sync.
3) This call should then just be:
`await this._roleManager.RefreshDosAdminRolesAsync(group.Users);`

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -228,13 +243,21 @@ public async Task<Group> AddChildGroups(Group group, IEnumerable<string> childGr
...
         }
 
         public async Task<Group> RemoveChildGroups(Group group, IEnumerable<string> childGroupNames)
         {
             var childGroups = await _groupStore.Get(childGroupNames);
-            return await _groupStore.RemoveChildGroups(group, childGroups);
+            var childUsers = childGroups.SelectMany(g => g.Users);
+
+            var persistedGroup = await _groupStore.RemoveChildGroups(group, childGroups);
+            await this._roleManager.RefreshDosAdminRolesAsync(childUsers);
```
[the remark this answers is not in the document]
bsmithvt: Similar to above,
1) ChildGroup navigation collection does not include Users, but this is irrelevant since AD groups cannot have users.
2) Remove `var childUsers = childGroups.SelectMany(g => g.Users);`
3) This call should then just be:
`await this._roleManager.RefreshDosAdminRolesAsync(group.Users);`

Fabric.Authorization.API/DependencyInjection/BaseSqlServerConfigurator.cs
hc-jordenlowe: Thats good to know, thanks.  We can remove it then.

Fabric.Authorization.API/DependencyInjection/InMemoryConfigurator.cs
hc-jordenlowe: We thought that.  We will tackle this after we get the app working as a whole.

Fabric.Authorization.Domain/Services/EDWAdminRoleSyncService.cs
hc-jordenlowe: We started looking at the database calls to make sure we need to make them.

Fabric.Authorization.API/DependencyInjection/InMemoryConfigurator.cs
hc-jordenlowe: I have wired up one.

Fabric.Authorization.Domain/Services/EDWAdminRoleSyncService.cs
hc-jordenlowe: Hey @michaelvidal24 , we have added feature flag and are ready to complete this pull request

PR #320, opened 2018-08-24
User/bsmithvt/access ctrl custom groups

bsmithvt:


Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
```diff
@@ -65,16 +67,25 @@ <h3>Selected</h3>
             <div class="table-container">
                 <table class="hc-table hc-table-borders">
                     <tbody>
-                        <tr *ngFor="let user of associatedPrincipals">
+                        <tr *ngFor="let user of associatedUsers">
```
michaelvidal24: Would it be better to collapse these into an associatedPrincipals? I see that is how it was before, just curious if we ran into issues with that, because I think the member component works similarly.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -65,18 +67,18 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
     this.editMode = !!this.groupName;
 
     if (this.editMode) {
-      Observable.zip(this.getGroupRoles(), this.getGroupUsers())
-        .do((result: [IRole[], IUser[]]) => {
+      Observable.zip(this.getGroupRoles(), this.getGroupUsers(), this.getChildGroups())
+        .do((result: [IRole[], IUser[], IGroup[]]) => {
           this.roles = result[0];
-          this.associatedPrincipals = result[1];
+          this.associatedUsers = result[1];
```
michaelvidal24: Similar comment as above, could we combine the results from the two calls into on associatedPrincipals array?

PR #326, opened 2018-08-29
User/jorden/synctests

hc-jordenlowe:
These tests will test the following scenarios:

- Remove datamartAdmin role to group, then all user should not be edwAdmin
- Remove dosAdmin role to group, then all user should not be edwAdmin
- Remove jobAdmin role to group, then all user should not be edwAdmin
- Check if user is not null
- Check if Ienumerable of user is not null
- If a user is deleted, then that user should be removed from edwAdmin
- If a group is deleted, then that group should be removed from edwAdmin
- If a child group (with role) is added to a parent group, and the user is in the parent group then the user should added
- Adding an admin role to child group should add it to a user in the parent group.
- Removing an admin role from a child group should remove it from a user in the parent group.


Fabric.Authorization.UnitTests/EDWAdminSyncService/EDWAdminRoleSyncServiceTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+        };
+
+        public static IEnumerable<object[]> DosAdminUserRoles => new []
+        {
+            new object[] 
+            {
+                new User("dosadminrole", "fabric-authorization")
+                {
+                    Roles = new Role[] { new Role { Name = "notadmin" }, new Role{ Name = "dosadmin" } }
```
michaelvidal24: In 18.2 we no longer have a dosadmin role - the roles we have that map to EDWAdmin are:

- datamartadmin
- jobadmin

DosAdmins is a custom group that has the datamartadmin, jobadmin and valueset-publisher roles.

Fabric.Authorization.Domain/Services/EDWAdminRoleSyncService.cs
```diff
@@ -66,14 +78,14 @@ private async Task<bool> IsUserASuperAdmin(User user)
             // if a group is a special Super Admin Group then return true
             // if not, then look on each group to see if there is a "super admin role"
             // finally, look on each group's children groups to see if there is a "super admin role"
-            foreach (var customGroup in user.Groups)
+            foreach (var customGroup in user.Groups.Where(p=> !p.IsDeleted))
```
bsmithvt: Are we returning deleted groups in the query that retrieves this user? That probably needs to be fixed at some point if so.

Fabric.Authorization.Domain/Services/EDWAdminRoleSyncService.cs
```diff
@@ -66,14 +78,14 @@ private async Task<bool> IsUserASuperAdmin(User user)
...
             {
                 if (customGroup.Roles.Any(r => RoleManagerConstants.AdminRoleNames.Contains(r.Name)))
                 {
                     return true;
                 }
 
-                if (customGroup.Children.SelectMany(children => children.Roles)
+                if (customGroup.Children.Where(group => !group.IsDeleted).SelectMany(children => children.Roles)
```
bsmithvt: Same question as above - I'm fairly certain this shouldn't be needed, but if it is, the query used to get the group should be adjusted.

michaelvidal24:
Here are some scenarios that I thought through that we should make sure we capture for testing:

Role associations
- user gets added to datamartadmin role -> add to edwadmin role
- user gets added to jobadmin role -> add to edwadmin role
- user gets added to a custom group that has the datamartadmin role -> add to edwadmin
- user gets added to a custom group that has the jobadmin role -> add to edwadmin
- datamartadmin role gets added to a custom group -> add all users in group to edwadmin
- jobadmin role gets added to a custom group -> add all users in group to edwadmin

Custom Group associations
- user gets removed from datamartadmin role -> remove user from edwadmin role if user doesn't have jobadmin role or datamartadmin role from another custom group
- user gets removed from jobadmin role -> remove user from edwadmin role if user doesn't have datamart admin role or jobadmin role from another custom group
- user gets removed from custom group that has the datamartadmin role -> remove user from edwadmin role if user doesn't have jobadmin role or datamartadmin role from another custom group
- user gets removed from custom group that has the jobadmin role -> remove user from edwadmin role if user doesn't have datamartadmin role or jobadmin role from another custom group
- datamartadmin role gets removed from custom group -> remove all users in group from edwadmin role if each user doesnt have jobadmin role or datamartadmin role from another custom group
- jobadmin role gets removed from custom group -> remove all users in group from edwadmin role if each user doesnt have datamartadmin role or jobadmin role from another custom group

Fabric.Authorization.UnitTests/EDWAdminSyncService/EDWAdminRoleSyncServiceTests.cs
hc-jordenlowe: cool, i have removed it

hc-jordenlowe:
Hey mike, thanks for adding some scenarios,  jason and I are adding the scenarios that follow:  

1. If adding a role and the user already have the edwAdmin, then do nothing
2. If removing a role and the user doesnt have the edwAdmin, then do nothing
3. If removing 1 role and a 2nd role has (one of the admin role), then do nothing
4. Add datamartAdmin role to user, then that user should be edwAdmin
5. Add dosAdmin role to user, then that user should be edwAdmin
6. Add jobAdmin role to user, then that user should be edwAdmin
7. Remove datamartAdmin role to user, then that user should not be edwAdmin
8. Remove dosAdmin role to user, then that user should not be edwAdmin
9. Remove jobAdmin role to user, then that user should not be edwAdmin
10. Add datamartAdmin role to group,  all users should be edwAdmin
11. Add dosAdmin role to group,  all users should be edwAdmin
12. Add jobAdmin role to group,  all users should be edwAdmin
13. Remove datamartAdmin role to group, then all user should not be edwAdmin
14. Remove dosAdmin role to group, then all user should not be edwAdmin
15. Remove jobAdmin role to group, then all user should not be edwAdmin
16. Check if user is not null
17. Check if Ienumerable of user is not null
18. If a user is deleted, then that user should be removed from edwAdmin
19. If a group is deleted, then that group should be removed from edwAdmin
20. if adding a role and all but 1 user has permission, then update that one user
21. if removing a role and all but 1 user has admin role, then remove from that one user
22. If a child group (with role) is added to a parent group, and the user is in the parent group then the user should added
23. If a child group (with role) is removed from a parent group, and the user is in the parent group then the user should removed
24. Adding an admin role to child group should add it to a user in the parent group.
25. Removing an admin role from a child group should remove it from a user in the parent group.

There are a few you mention that we will go back and add.  However some of these I believe will be acceptance/integration/functional tests based on their nature, ill update which ones we need in high level testing

michaelvidal24:
Sounds good - thanks for the comprehensive thought on this feature.

PR #329, opened 2018-08-30
User/bsmithvt/manageauth perm save

bsmithvt:


Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  private permissions: Array<string> = [];
+  private initialized: boolean;
+
+  constructor(private authUserService: FabricAuthUserService) {
+  }
+
+  getPermissions(): Observable<string[]> {
+    if (!this.initialized) {
+      this.initialized = true;
+      return this.authUserService.getCurrentUserPermissions().do(userPermissionResponse => {
```
michaelvidal24: Should we introduce a caching window? I don't think we want to cache these permissions forever...

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -58,15 +60,20 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
...
   ) { }
 
   ngOnInit() {
+
     this.groupName = this.route.snapshot.paramMap.get('subjectid');
     this.grain = this.route.snapshot.paramMap.get('grain');
     this.securableItem = this.route.snapshot.paramMap.get('securableItem');
     this.editMode = !!this.groupName;
+    this.currentUserService.getPermissions().subscribe(p => {
+      this.missingManageAuthorizationPermission = !p.includes(`${this.grain}/${this.securableItem}.manageauthorization`);
```
michaelvidal24: Should this check also check the roles associated to the custom group, if any? Since a custom group can be associated to roles across securables, we will deny the permission on the server if the user doesn't have manageauthorization permission for the securables tied to all the roles...

Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
michaelvidal24: perhaps 2 minutes...

PR #332, opened 2018-08-31
User/jason/edwadminmoduletests

hc-jordenlowe:
Added some test scenarios:

1.  If you try to sync the wrong group, it will throw 404
2.  if you add or remove a role from a user, it will add remove the EDW Admin role
3.  If you add or remove a role from a group, it will add remove the EDW Admin role from the user
4.   If you add or remove a user from a group, it will add remove the EDW Admin role from the user
5.  If you remove a user with a role jobsadmin from a group with jobs admin, then the user will stay in edw admin.


Fabric.Authorization.API/Modules/EdwAdminModule.cs
```diff
@@ -67,6 +67,10 @@ private async Task<dynamic> SyncGroupIdentityRoles(dynamic param)
                 }
                 return HttpStatusCode.NoContent;
             }
+            catch (NotFoundException<Group>)
+            {
+                return CreateFailureResponse($"TEMP Group not found", HttpStatusCode.NotFound);
```
michaelvidal24: Is TEMP a placeholder for something here?

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -363,6 +363,16 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
                   .then(value => {
                       return this.edwAdminService.syncGroupWithEdwAdmin(newGroup.groupName)
                           .toPromise().then(o => { return value; }).catch(err => { return value; });
+                  })
+                  .then(value => {
+                      if (usersToRemove) {
+                        let response;
+                        usersToRemove.forEach(element => {
+                          response = this.edwAdminService.syncUserWithEdwAdmin(element.subjectId, element.identityProvider)
```
michaelvidal24: this might get a little chatty if we are removing a large list of users...can we add an endpoint that takes a list of users instead?

hc-jordenlowe: ah, we can look over that for sure.   Maybe replace our single user and make it take it a list regardless.

PR #334, opened 2018-08-31
Fix endless spinner when grain is selected and attempt to order

ctediz:
Do not switch nodes if the node being selected is a parent grain node

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.spec.ts
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.spec.ts
@@ -180,7 +180,7 @@ describe('GrainListComponent', () => {
 
   it('onSelect sets the selectedNode', () => {
     // Arrange
-    const randomNode = new GrainFlatNode(true, 'test', null, 0);
+    const randomNode = new GrainFlatNode(true, 'test', 'parent', 0);
     component.selectedNode = null;
 
     // Act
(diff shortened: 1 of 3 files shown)


Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.ts
```diff
@@ -139,13 +139,11 @@ export class GrainListComponent implements OnInit {
   }
 
   onSelect(node: GrainFlatNode): void {
-    this.selectedNode = node;
-
     if (!!node.parentName) {
-      this.setSelectedNode(this.getSelectedGrain(), node.name);
-    } else {
-      this.setSelectedNode(node.name);
+      this.selectedNode = node;
```
bsmithvt: What happens now if the user clicks on the grain (node without a parent)?

Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.spec.ts
```diff
@@ -72,7 +72,7 @@ describe('GrainListComponent', () => {
 
     // assert
     expect(component.selectedNode).not.toEqual(null);
-    expect(component.selectedNode.name).toBe(grains[0].name);
+    expect(component.selectedNode.name).toBe(grains[0].securableItems[0].name);
```
ctediz: The grains-list will default to the first securable item when none is explicitly passed in

PR #338, opened 2018-09-05
ensure url includes grain and securable item when selecting grain or …

bsmithvt:
…securable to fix deep linking bug

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.ts
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.ts
@@ -63,7 +63,6 @@ export class GrainListComponent implements OnInit {
   }
 
   ngOnInit() {
-
   }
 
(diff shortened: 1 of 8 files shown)


Fabric.Authorization.AccessControl/src/app/modules/access-control/grain-list/grain-list.component.ts
```diff
@@ -155,12 +154,16 @@ export class GrainListComponent implements OnInit {
   }
 
   setSelectedNode(grain: string, securableItem?: string) {
+    const location = window.location;
+    const url = `${location.protocol}//${location.host}/client/access-control`;
```
hc-jordenlowe: will host have the port number associated with it?  just making sure its parsing correctly

bsmithvt: Yes `host` includes the port. I'll try to pull it from config.

PR #343, opened 2018-09-07
Show role.name when role.displaName is null.

michaelvidal24:


bsmithvt:
Should we display some default text for null description or just leave it blank?

michaelvidal24:
I think it's probably fine to leave it blank.

PR #358, opened 2018-09-17
IE no-cache required headers

ctediz:
Do not cache any requests made from access-control


ctediz:
Data-atlas has code to not cache only calls to mds.  Is it safe to not cache any calls in general, or should we take a similar approach and not cache from sites we specify? (which currently would only be auth/identity i believe)

johnrulon:
I think it would be best to only apply these for IE if we can.

PR #366, opened 2018-10-05
Case insensitive check for existing group

ctediz:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
```diff
@@ -457,7 +457,7 @@ public Task<IEnumerable<Group>> Get(IEnumerable<string> groupNames)
                  && groupNames.Contains(g.Name)
                  );
 
-            var missingGroups = groupNames.Except(groupEntities.Select(g => g.Name)).ToList();
+            var missingGroups = groupNames.Except(groupEntities.Select(g => g.Name), StringComparer.OrdinalIgnoreCase).ToList();
```
hc-jordenlowe: Is this something where you can add an integration test to make sure the ignore case works?

Add something like "Group 1" and search for "group 1"?

michaelvidal24: I wonder if there is a better place to put this check? Seems like we want to prevent duplicate group names from getting into the data store, so maybe we need to look at where those groups are being added and add better validation there? Validation seems more like business logic to me, so I feel like that should go in our domain layer where other validation logic resides.

ctediz: We decided to leave this for now, going to look into writing a test for this

PR #373, opened 2018-10-11
1) decrypt config settings if encrypted, 2) set user display name

bsmithvt:


bsmithvt:
@michaelvidal24 ah, good catch. Thanks.

PR #382, opened 2018-10-22
Added script to install node, npm, angular and more.

hc-jordenlowe:
Installs and uninstalls the following:

1. node modules - this is to get rid of specific platform dependencies
2. node
3. angular
4. typescript
5. npm install

There may be one more step left.  I noticed node-sass needs to be precompiled before you can use access-control.  So i may need to fix that up some.

### DIFF
--- a/Fabric.Authorization.AccessControl/setup-dependencies.ps1
+++ b/Fabric.Authorization.AccessControl/setup-dependencies.ps1
@@ -3,11 +3,18 @@
 # setup-dependencies.ps1
 #
 
+# Change these values as needed.
 $nodeVersion = "8.12.0"
 $angularVersion = "6.1.2"
 $typescriptVersion = "2.9.2"
 $npmVersion = "6.4.1"
 $nl = [Environment]::NewLine
+
+# This will be the log for any application having issues
+$Logfile = ".\setup-dependencies.log"
+
+# This is to throw exceptions for any app that has issues.
+$ErrorActionPreference = "Stop"
 
 # Chocolatey must be installed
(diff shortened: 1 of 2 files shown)


hc-jordenlowe:
So npm comes with nodejs pre-pacakged, but we may want to use a lower version of an NPM than nodejs.  Ill add a line to up/downgrade npm

Fabric.Authorization.AccessControl/setup-dependencies.ps1
```diff
@@ -0,0 +1,40 @@
...
+$typescriptVersion = "3.1.3"
+
+# remove node_modules
+Write-Output ""
+Write-Output "$nl----] Removing previous node modules [----$nl"
+Remove-Item .\\node_modules -Force -Recurse
+
+# uninstall angular/cli
+Write-Output "$nl----] Removing previous angular cli [----$nl"
+npm uninstall -g @angular/cli
```
ryanorbaker: should we check for the existence of choco before we delete?  if for some reason choco is not installed and we remove things, how do we put them back?

hc-jordenlowe:
Added exception handling so it goes to a log file and makes the commandline less chaotic.  Added a check at the end to show the versions to make sure things were updated.  Also added some messaging to the Readme file

Fabric.Authorization.AccessControl/setup-dependencies.ps1
hc-jordenlowe: All set above.  I even install it for you too!

PR #383, opened 2018-10-24
User/bsmithvt/idpss integration tests

bsmithvt:


Fabric.Authorization.API/RemoteServices/IdentityProviderSearch/Providers/IIdPSearchProvider.cs
```diff
@@ -6,5 +6,6 @@ namespace Fabric.Authorization.API.RemoteServices.IdentityProviderSearch.Provide
     public interface IIdPSearchProvider
     {
         Task<FabricIdPSearchResponse> Search(IdPPrincipalSearchRequest request);
+        Task<FabricIdPGroupResponse> GetGroup(IdPGroupRequest request);
```
hc-jordenlowe: probably rename this GetGroupAsync as well as the SearchAsync if they are

PR #385, opened 2018-10-25
Added function to add user to edwAdmin

hc-jordenlowe:
Added SQL script to Add-UserOrGroupToEdwAdmin.  Double check my SQL and see if it sucks

Added function Add-AccountToEDWAdmin to add a user to EDWAdmin

Added function call right below adding the user to DosAdmin

I need to double check if the user exists to not add them

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
@@ -270,6 +270,41 @@ function Add-RoleToGroup
     Invoke-Sql $connString $query @{groupId = $groupId; roleId = $roleId; ; clientId = $clientId} | Out-Null
 }
 
+function Add-UserOrGroupToEdwAdmin($userOrGroup, $connString) {
+	$query   = "DECLARE @userId INTEGER;
+                DECLARE @roleId INTEGER;
+                
+                SET @userId = -1
+                SELECT @userId = COALESCE([IdentityID], -1) FROM [EDWAdmin].[CatalystAdmin].[IdentityBASE] WHERE [IdentityNM] = @identityName
+                BEGIN TRY
+                BEGIN TRANSACTION
+                    IF (@userId < 0)
+                    BEGIN
+                        INSERT INTO [EDWAdmin].[CatalystAdmin].[IdentityBASE] ([IdentityNM])
+                        VALUES (@identityName);
+                        SELECT @userId = SCOPE_IDENTITY();
+                    END
+                
+                    SELECT @roleId = [RoleID]
+                    FROM [EDWAdmin].[CatalystAdmin].[RoleBASE]
+                    WHERE [RoleNM] = @roleName;
+                
+                    IF NOT EXISTS (SELECT * FROM [EDWAdmin].[CatalystAdmin].[IdentityRoleBASE] WHERE [IdentityID] = @userId AND [RoleId] = @roleId)
+                    BEGIN
+                        INSERT INTO [EDWAdmin].[CatalystAdmin].[IdentityRoleBASE] (IdentityID, RoleID)
+                        VALUES(@userId, @roleId);
+                    END
+                        
+                    COMMIT;
+                END TRY
+                BEGIN CATCH
+                    ROLLBACK;
+                    THROW; 
+                END CATCH"
+                
+    Invoke-Sql -connectionString $connString -sql $query -parameters @{roleName = "EDW Admin"; identityName = $userOrGroup} | Out-Null
+}
+
 function Get-PrincipalContext
 {
     param(
@@ -383,6 +418,17 @@ function Add-ListOfUsersToDosAdminGroup($edwAdminUsers, $connString, $authorizat
             }
         }        
     }	
+}
+
+function Add-AccountToEDWAdmin($accountName, $domain, $connString) {
+	$samAccountName = Get-SamAccountFromAccountName -accountName $accountName
+	if ((Test-IsUser -samAccountName $samAccountName -domain $domain) -or (Test-IsGroup -samAccountName $samAccountName -domain $domain)) {
+        Add-UserOrGroupToEdwAdmin -userOrGroup $accountName -connString $connString
+	}
+    else {
+        Write-DosMessage "$samAccountName is not a valid principal in the $domain domain. Please enter a valid account. Halting installation."
+        throw
+    }
 }
 
 function Add-DosAdminGroup
@@ -1212,6 +1258,7 @@ Export-ModuleMember Get-AdminAccount
 Export-ModuleMember Set-AuthorizationEnvironmentVariables
 Export-ModuleMember Move-DosAdminRoleToDosAdminGroup
 Export-ModuleMember Add-AccountToDosAdminGroup
+Export-ModuleMember Add-AccountToEDWAdmin
 Export-ModuleMember Invoke-MonitorShallow
 Export-ModuleMember Add-EdwAdminUsersToDosAdminGroup
 Export-ModuleMember Add-AuthorizationRegistration
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Utilities.tests.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Utilities.tests.ps1
@@ -1443,3 +1443,66 @@ Describe 'Add-AccountToDosAdminGroup tests' -Tag 'Unit'{
         }
     }
 }
+
+Describe 'Test-Add-AccountToEDWAdmin tests' -Tag 'Unit' {
+    InModuleScope Install-Authorization-Utilities{
+        It 'Should throw error if not a user or group' {
+            # Arrange
+            Mock Test-IsUser { return $false }
+            Mock Test-IsGroup { return $false }
+            Mock Write-DosMessage {}
+
+            # Act and Assert
+            { Add-AccountToEDWAdmin -accountName "test\user" -domain "domain" -connString "connString" } | Should -Throw
+        }
+
+        It 'Should throw error if domain\user is incorrect' {
+            # Arrange
+            Mock Test-IsUser { return $true }
+            Mock Test-IsGroup { return $false }
+
+            # Act and Assert
+            { Add-AccountToEDWAdmin -accountName "test.user" -domain "domain" -connString "connString" } | Should -Throw
+        }
+
+        It 'Should work if accountName is user' {
+            # Arrange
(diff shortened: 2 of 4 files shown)


Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -147,6 +147,36 @@ function Add-User($authUrl, $name, $accessToken) {
     return Invoke-Post $url $body $accessToken
 }
 
+function Add-UserOrGroupToEdwAdmin($userOrGroup, $connString) {
+	$query = "DECLARE @userId INTEGER;
+			  DECLARE @roleId INTEGER;
+			  IF NOT EXISTS (SELECT * FROM [EDWAdmin].[CatalystAdmin].[IdentityBASE] WHERE [IdentityNM] = @identityName)
```
michaelvidal24: Do we need to handle adding the user if the user doesn't already exist in IdentityBASE?

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -385,6 +411,27 @@ function Add-ListOfUsersToDosAdminGroup($edwAdminUsers, $connString, $authorizat
...
+function Add-AccountToEDWAdmin($accountName, $domain, $connString) {
+	$samAccountName = Get-SamAccountFromAccountName -accountName $accountName
+	if ((Test-IsUser -samAccountName $samAccountName -domain $domain) -or (Test-IsGroup -samAccountName $samAccountName -domain $domain)) {
+		try {
+			
+			Add-UserOrGroupToEdwAdmin -userOrGroup $accountName -connString $connString
+		}
+		catch {
+            $exception = $_.Exception
+            if ($exception.Response -ne $null) {
```
michaelvidal24: The exception in this case will never have a Response object, that only exists on System.Net.WebException I believe, which won't be thrown here.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
hc-jordenlowe: thats what this line is for:

IF NOT EXISTS (SELECT * FROM [EDWAdmin].[CatalystAdmin].[IdentityBASE] WHERE [IdentityNM] = @identityName)

however, if the user exists and they do not have permission, then I need to handle this case

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
hc-jordenlowe: Ah, good catch.  I was wondering why the exceptions didnt make sense

hc-jordenlowe: Removed the exception altogether.  No need to catch it.

It was a case of copy pasta

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
hc-jordenlowe: Tested this functionality.  It will insert the user if its not there, then it will check to see if that user has access and will add it if they do not.

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -270,6 +270,43 @@ function Add-RoleToGroup
     Invoke-Sql $connString $query @{groupId = $groupId; roleId = $roleId; ; clientId = $clientId} | Out-Null
 }
 
+function Add-UserOrGroupToEdwAdmin($userOrGroup, $connString) {
+	$query   = "DECLARE @userId INTEGER;
+                DECLARE @roleId INTEGER;
+                
+                SET @userId = -1
+                SELECT @userId = COALESCE([IdentityID], -1) FROM [EDWAdmin].[CatalystAdmin].[IdentityBASE] WHERE [IdentityNM] = @identityName
```
hckenmiller: this is a global variable, right?

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -270,6 +270,43 @@ function Add-RoleToGroup
...
+                        INSERT INTO [EDWAdmin].[CatalystAdmin].[IdentityBASE] ([IdentityNM])
+                        VALUES (@identityName);
+                        SELECT @userId = SCOPE_IDENTITY();
+                    END
+                
+                    SELECT @roleId = [RoleID]
+                    FROM [EDWAdmin].[CatalystAdmin].[RoleBASE]
+                    WHERE [RoleNM] = @roleName;
+                
+                    SELECT @userId, @roleId;
```
hckenmiller: This Select is probably only to test the values, remove.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -147,6 +147,44 @@ function Add-User($authUrl, $name, $accessToken) {
     return Invoke-Post $url $body $accessToken
 }
 
+function Add-UserOrGroupToEdwAdmin($userOrGroup, $connString) {
+	$query   = "DECLARE @userId INTEGER;
```
ryanorbaker: why is this here as well as utilities?  
I must be missing something?

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -270,6 +270,43 @@ function Add-RoleToGroup
...
+                    IF (@userId < 0)
+                    BEGIN
+                        INSERT INTO [EDWAdmin].[CatalystAdmin].[IdentityBASE] ([IdentityNM])
+                        VALUES (@identityName);
+                        SELECT @userId = SCOPE_IDENTITY();
+                    END
+                
+                    SELECT @roleId = [RoleID]
+                    FROM [EDWAdmin].[CatalystAdmin].[RoleBASE]
+                    WHERE [RoleNM] = @roleName;
```
hckenmiller: global?

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -270,6 +270,43 @@ function Add-RoleToGroup
...
+                    END
+                        
+                    COMMIT;
+                END TRY
+                BEGIN CATCH
+                    ROLLBACK;
+                    THROW; 
+                END CATCH"
+                
+    Invoke-Sql -connectionString $connString -sql $query -parameters @{roleName = "EDW Admin"; identityName = $userOrGroup} | Out-Null
```
hckenmiller: global? userorgroup

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.tests.ps1
```diff
@@ -1443,3 +1443,66 @@ Describe 'Add-AccountToDosAdminGroup tests' -Tag 'Unit'{
...
+            Mock Test-IsGroup { return $false }
+            Mock Write-DosMessage {}
+
+            # Act and Assert
+            { Add-AccountToEDWAdmin -accountName "test\user" -domain "domain" -connString "connString" } | Should -Throw
+        }
+
+        It 'Should throw error if domain\user is incorrect' {
+            # Arrange
+            Mock Test-IsUser { return $True }
```
hckenmiller: lowercase to match all the others $true

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
```diff
@@ -147,6 +147,44 @@ function Add-User($authUrl, $name, $accessToken) {
...
+                        INSERT INTO [EDWAdmin].[CatalystAdmin].[IdentityBASE] ([IdentityNM])
+                        VALUES (@identityName);
+                        SELECT @userId = SCOPE_IDENTITY();
+                    END
+                
+                    SELECT @roleId = [RoleID]
+                    FROM [EDWAdmin].[CatalystAdmin].[RoleBASE]
+                    WHERE [RoleNM] = @roleName;
+                
+                    SELECT @userId, @roleId;
```
hckenmiller: remove it only is to test the values

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
hc-jordenlowe: That is a sql parameter variable.

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
hc-jordenlowe: So if you look, there are two files with the same functions.  That is because one is a Utilities powershell script and the other is this, which is the older powershell functions.  Anyways, as I understand the older powershell script, this one, is depreciated and will be removed eventually and replaced by the utility script and the quiet install.

So if it is going to be removed soon, do we want to reuse the code or just let the old code die?

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
hc-jordenlowe: This is a sql parameter name

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
hckenmiller: I'm not seeing where it gets it's value set

Fabric.Authorization.API/scripts/Install-Authorization-Windows.ps1
michaelvidal24: After this change, I think we can let the old code die. No need to do additional work to include the utilities module here.

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -270,6 +270,41 @@ function Add-RoleToGroup
...
```
hckenmiller: This is where roleName and identity are assigned.

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
hckenmiller: Yah I figure it out by looking at the invoke where the -parameters are declared/defined.

PR #387, opened 2018-10-30
User/bsmithvt/access control aad

bsmithvt:


Fabric.Authorization.Domain/Validators/GroupValidator.cs
```diff
@@ -23,10 +23,10 @@ private void ConfigureRules()
                 .WithMessage("Please specify a Name for this Group.")
                 .WithState(g => ValidationEnums.ValidationState.MissingRequiredField);
 
-            RuleFor(group => group.IdentityProvider)
+            /*RuleFor(group => group.IdentityProvider)
```
ctediz: Is this meant to be here ?

Fabric.Authorization.API/Modules/GroupsModule.cs
```diff
@@ -162,10 +161,12 @@ private async Task<dynamic> GetRolesForGroup(dynamic p)
             try
             {
                 this.RequiresClaims(AuthorizationReadClaim);
-                string groupName = p.groupName;
                 string grain = p.grain;
                 string securableItem = p.securableItem;
-                var group = await _groupService.GetGroup(groupName);
+                var groupIdentifierApiRequest = this.Bind<GroupIdentifierApiRequest>();
```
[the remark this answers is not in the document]
bsmithvt: `Bind` is supposed to bind the path and query parameters to the model assuming the model property names match the query parameter key / the path parameter variable defined in the routes. I did hit some issues with query parameters recently that I need to figure out.

Fabric.Authorization.API/RemoteServices/IdentityProviderSearch/Providers/IdPSearchProvider.cs
```diff
@@ -89,14 +112,16 @@ private async Task<HttpRequestMessage> CreateHttpRequestMessage(string route, Ht
...
                 accessTokenResponse.AccessToken);
 
+            httpRequestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
             return httpRequestMessage;
         }
 
-        private async Task<T1> ProcessResponse<T, T1>(HttpRequestMessage httpRequestMessage)
-            where T1 : IFabricIdPResponseModel<T>, new()
+        /*private async Task<T1> ProcessResponse<T, T1>(HttpRequestMessage httpRequestMessage)
```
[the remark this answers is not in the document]
bsmithvt: Yes, I had to make some adjustments today but want to revisit the generic method in the future.

Fabric.Authorization.API/appsettings.json
```diff
@@ -32,8 +32,8 @@
   },
   "DefaultPropertySettings": {
     "GroupSource": "Directory",
-    "DualStoreEDWAdminPermissions": "true",
-    "IdentityProvider":  "Windows" 
+    "DualStoreEDWAdminPermissions": "true"
+    // "IdentityProvider":  "Windows"
```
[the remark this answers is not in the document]
bsmithvt: Yes, this is used as the default IdP for groups. However, I need to temporarily comment out until appropriate installer changes can be made to update existing data so we don't break anyone.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.html
```diff
@@ -34,7 +34,7 @@ <h3>Available</h3>
                     <li *ngFor="let principal of principals">
                         <hc-checkbox [id]="principal.subjectId + principal.principalType" [(ngModel)]="principal.selected">
                         </hc-checkbox>
-                        <span [innerText]="principal.subjectId">
+                        <span [innerText]="principal.displayName">
```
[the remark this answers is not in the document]
bsmithvt: This info is coming back from the IdP SS. I will confirm that it's always being set in there.

Fabric.Authorization.Domain/Validators/GroupValidator.cs
bsmithvt: This is temporarily commented out for the same reason as the `appsettings.json` default IdP setting.

Fabric.Authorization.API/Modules/EdwAdminModule.cs
```diff
@@ -72,23 +74,34 @@ private async Task<dynamic> SyncGroupIdentityRoles(dynamic param)
         {
             CheckInternalAccess();
 
+            var groupIdentifier = new GroupIdentifier
+            {
+                GroupName = param.groupName,
+                IdentityProvider = SetIdentityProvider(param.identityProvider),
+                TenantId =  param.tenantId
```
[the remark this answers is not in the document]
bsmithvt: Yes, there are cases where IdP and TenantId will be null, whether because of existing clients not including that info in the request or in the case of AD, they will never have a TenantId. For the case where clients don't post the info in the request when creating the group, we default to the IdP specified in the config file.

Fabric.Authorization.API/RemoteServices/IdentityProviderSearch/Providers/IdPSearchProvider.cs
```diff
@@ -56,16 +58,16 @@ public async Task<FabricIdPSearchResponse> SearchAsync(IdPPrincipalSearchRequest
             httpRequestMessage.Content = new StringContent(JsonConvert.SerializeObject(request), Encoding.UTF8,
                 "application/json");
 
-            return await ProcessResponse<IdPPrincipalSearchResponse, FabricIdPSearchResponse>(httpRequestMessage);
+            return new FabricIdPSearchResponse();
```
michaelvidal24: I think I am missing something, but where do we invoke the actual search?

Fabric.Authorization.API/RemoteServices/IdentityProviderSearch/Providers/IdPSearchProvider.cs
```diff
@@ -75,7 +77,28 @@ public async Task<FabricIdPGroupResponse> GetGroupAsync(IdPGroupRequest request)
...
+            {
+                _logger.Error(
+                    $"Response status code from {ServiceName} {httpRequestMessage.RequestUri} => {response.StatusCode}");
+                _logger.Error(
+                    $"Response content from {ServiceName} {httpRequestMessage.RequestUri} => {responseContent}");
+            }
+            else
+            {
+                idPGroupResponse = JsonConvert.DeserializeObject<IdPGroupResponse>(responseContent);
+                _logger.Debug($"{ServiceName} {httpRequestMessage.RequestUri} result: {idPGroupResponse}");
```
michaelvidal24: I like the use of the debug log level here.

Fabric.Authorization.AccessControl/src/app/services/fabric-auth-edwadmin.service.ts
```diff
@@ -52,4 +53,22 @@ export class FabricAuthEdwAdminService extends FabricBaseService {
...
     );
   }
+
+  private setIdPAndTenant(url: string, identityProvider: string, tenantId: string) {
+    url = this.setQueryParameters(url, 'identityProvider', identityProvider);
+    url = this.setQueryParameters(url, 'tenantId', tenantId);
+    return url;
+  }
+
+  private setQueryParameters(url: string, key: string, val: string) {
```
michaelvidal24: Is there an easier way of building up a url with query params in angular than manually parsing the string?

Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerRoleStore.cs
```diff
@@ -233,53 +239,60 @@ private async Task<EntityModels.Role> GetEntityModel(Guid id)
...
-            {
-                roles = roles.Where(r => r.Name == roleName);
-            }
+                if (!string.IsNullOrWhiteSpace(roleName))
+                {
+                    roles = roles.Where(r => r.Name == roleName);
+                }
 
-            foreach (var role in roles)
+                foreach (var role in roles)
```
michaelvidal24: I know this was here before, but this loop over an IQueryable with ToLists inside of it seems to be a performance red flag...we don't have to fix it with this PR, but perhaps we need to address it in a subsequent PR.

Fabric.Authorization.API/RemoteServices/IdentityProviderSearch/Providers/IdPSearchProvider.cs
bsmithvt: We do not. We just use the other method in this class. I'll remove this to avoid confusion.

PR #389, opened 2018-11-02
ensure permissions for potential duplicate group names across IdPs/te…

bsmithvt:
…nants are resolved correctly

### DIFF
--- a/Fabric.Authorization.API/Modules/UsersModule.cs
+++ b/Fabric.Authorization.API/Modules/UsersModule.cs
@@ -134,7 +134,7 @@ namespace Fabric.Authorization.API.Modules
                 Grain = userPermissionRequest.Grain,
                 SecurableItem = userPermissionRequest.SecurableItem,
                 IncludeSharedPermissions = isGrainEmpty,
-                UserGroups = groups?.Select(g => g.Name)
+                UserGroups = groups
             });
 
             return permissionResolutionResult.AllowedPermissions
--- a/Fabric.Authorization.Domain/Resolvers/Permissions/RolePermissionResolverService.cs
+++ b/Fabric.Authorization.Domain/Resolvers/Permissions/RolePermissionResolverService.cs
@@ -21,7 +21,7 @@ namespace Fabric.Authorization.Domain.Resolvers.Permissions
         {
             var grain = resolutionRequest.Grain;
             var securableItem = resolutionRequest.SecurableItem;
-            var groupNames = resolutionRequest.UserGroups.ToList();
+            var groups = resolutionRequest.UserGroups.ToList();
 
             var roles = (await _roleService.GetRoles(grain, securableItem)).ToList();
 
@@ -30,7 +30,7 @@ namespace Fabric.Authorization.Domain.Resolvers.Permissions
 
             foreach (var role in roles)
             {
-                if (BypassRoleEvaluation(role, groupNames, resolutionRequest))
+                if (BypassRoleEvaluation(role, groups, resolutionRequest))
                 {
                     continue;
                 }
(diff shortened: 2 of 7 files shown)


ryanorbaker:
for my own understanding:  if I understand the change, we are switching from passing group names around to passing group objects (with names and ids) so we can duplicate names?

Fabric.Authorization.Domain/Resolvers/Permissions/RolePermissionResolverService.cs
```diff
@@ -78,14 +78,16 @@ public async Task<PermissionResolutionResult> Resolve(PermissionResolutionReques
...
-        private static bool BypassRoleEvaluation(Role role, IReadOnlyCollection<string> groupNames, PermissionResolutionRequest permissionResolutionRequest)
+        private static bool BypassRoleEvaluation(Role role, IReadOnlyCollection<Group> groups, PermissionResolutionRequest permissionResolutionRequest)
         {
-            return !(role.Groups.Any(g => groupNames.Contains(g.Name, StringComparer.OrdinalIgnoreCase)) || role.Users.Any(
-                         u => u.SubjectId.Equals(permissionResolutionRequest.SubjectId,
-                                  StringComparison.OrdinalIgnoreCase) &&
-                              u.IdentityProvider.Equals(permissionResolutionRequest.IdentityProvider,
-                                  StringComparison.OrdinalIgnoreCase))) || role.IsDeleted ||
-                   role.Permissions == null || role.Permissions.Count == 0;
+            return !(role.Groups.Any(g => groups.Select(g1 => g1.GroupIdentifier).Contains(g.GroupIdentifier, new GroupIdentifierComparer()))
```
ryanorbaker: is this where you think we need to use the *correct* ID, either from Azure or other?

bsmithvt:
@ryanorbaker, yes, since with Azure AD we can have duplicate group names that span tenants, the new group identifier will be the IdP + tenant + group name.

Fabric.Authorization.API/Modules/UsersModule.cs
```diff
@@ -134,7 +134,7 @@ private async Task<dynamic> GetUserPermissions(dynamic param)
                 Grain = userPermissionRequest.Grain,
                 SecurableItem = userPermissionRequest.SecurableItem,
                 IncludeSharedPermissions = isGrainEmpty,
-                UserGroups = groups?.Select(g => g.Name)
+                UserGroups = groups
```
michaelvidal24: Do we need a similar change here: https://github.com/HealthCatalyst/Fabric.Authorization/pull/389/files#diff-8aa3fecd22b341d99ca92ccc12c0f59dR159, or is that already handled?

Fabric.Authorization.Domain/Resolvers/Permissions/RolePermissionResolverService.cs
bsmithvt: The GroupIdentifier is a property consisting of IdP + tenant ID + group name. This logic was switched from comparing group names to comparing `GroupIdentifier` objects.

bsmithvt: The Azure identifier is used to correlate internal Fabric.Auth groups to Azure groups. We correlate through the new `ExternalIdentifier` column. Access tokens from Azure AD will have that in the `groups` claim so we have to translate internally. This is done up the call stack from this method.

Fabric.Authorization.API/Modules/UsersModule.cs
bsmithvt: Not sure what you mean by "similar change", but this endpoint should be handled. I switched the `PermissionResolutionRequest` from having a list of group names to having a list of `GroupIdentifier` objects.

michaelvidal24: Just wanting to make sure the GetCurrentUserPermissions has been updated to reflect UserGroups changing from being a list of `string` to a list of `GroupIdentifier` objects.

bsmithvt: Ah yeah, it has been.

PR #391, opened 2018-11-16
User/bsmithvt/group name casing fix

bsmithvt:
I have some more integration test work to do but the bulk of the logic is done and can be reviewed.

### DIFF
--- a/Fabric.Authorization.API/Bootstrapper.cs
+++ b/Fabric.Authorization.API/Bootstrapper.cs
@@ -107,6 +107,9 @@ namespace Fabric.Authorization.API
             ConfigureApplicationRegistrations(container);
             var dbBootstrapper = container.Resolve<IDbBootstrapper>();
             dbBootstrapper.Setup();
+
+            var appConfig = container.Resolve<IAppConfiguration>();
+            if (appConfig.MigrateDuplicateGroups) { }
         }
 
         private void InitializeSwaggerMetadata()
@@ -149,6 +152,7 @@ namespace Fabric.Authorization.API
             container.Register(typeof(IOptions<>), typeof(OptionsManager<>));
             container.Register<IMemoryCache, MemoryCache>();
             container.Register<Domain.Defaults.Authorization>();
+            container.Register<GroupMigratorService, GroupMigratorService>();
 
             var httpClient = new HttpClient();
             container.Register(httpClient);
@@ -170,6 +174,13 @@ namespace Fabric.Authorization.API
 
             var configurator = container.Resolve<IPersistenceConfigurator>();
             configurator.ConfigureApplicationInstances(container);
+
+            var appConfig = container.Resolve<IAppConfiguration>();
+            if (appConfig.MigrateDuplicateGroups)
+            {
+                var groupMigratorService = container.Resolve<GroupMigratorService>();
+                groupMigratorService.MigrateDuplicateGroups();
+            }
         }
 
         protected override void ConfigureConventions(NancyConventions nancyConventions)
--- a/Fabric.Authorization.API/Bootstrapper.cs
+++ b/Fabric.Authorization.API/Bootstrapper.cs
@@ -76,7 +76,7 @@ namespace Fabric.Authorization.API
             configurator.ConfigureRequestInstances(container);
         }
         
-        protected override void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
+        protected override async void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
         {
             TinyIoCContainer = container;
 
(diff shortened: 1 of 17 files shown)


Fabric.Authorization.API/Bootstrapper.cs
```diff
@@ -107,6 +107,13 @@ protected override void ApplicationStartup(TinyIoCContainer container, IPipeline
             ConfigureApplicationRegistrations(container);
             var dbBootstrapper = container.Resolve<IDbBootstrapper>();
             dbBootstrapper.Setup();
+
+            var appConfig = container.Resolve<IAppConfiguration>();
+            if (appConfig.MigrateDuplicateGroups)
+            {
+                var groupMigratorService = container.Resolve<GroupMigratorService>();
+                var migrationResults = await groupMigratorService.MigrateDuplicateGroups();
```
michaelvidal24: What do we do with this result?

Fabric.Authorization.API/appsettings.json
```diff
@@ -46,5 +46,6 @@
   },
   "EntityFrameworkSettings": {
     "MinimumLogLevel": "Warning"
-  }
+  },
+  "MigrateDuplicateGroups": true
```
michaelvidal24: Should we also add a default to the web.config, that way we can tell a user to flip it to false if necessary.

Fabric.Authorization.Domain/Services/GroupMigratorService.cs
```diff
@@ -0,0 +1,156 @@
...
+
+                    // migrate directory groups to new parents (this will not execute if the duplicate group is a custom group)
+                    foreach (var parent in duplicateGroup.Parents)
+                    {
+                        var parentExists = originalGroup.Parents.Any(p => p.Id == parent.Id);
+                        if (!parentExists)
+                        {
+                            _logger.Information($"Migrating Group {originalGroup} to Parent Group {parent}");
+                            originalGroup.Parents.Add(parent);
+                            /*
```
michaelvidal24: Should we get rid of this commented code?

Fabric.Authorization.API/Bootstrapper.cs
bsmithvt: Nothing yet - I was thinking of potentially logging the entire structure out.

Fabric.Authorization.API/appsettings.json
bsmithvt: Yes

Fabric.Authorization.Domain/Services/GroupMigratorService.cs
bsmithvt: Yes, I was just waiting to make sure the AutoMapper profile stuff was going to work out.

Fabric.Authorization.Domain/Services/GroupMigratorService.cs
```diff
@@ -0,0 +1,158 @@
...
+            IGroupStore groupStore,
+            ILogger logger)
+        {
+            _groupStore = groupStore ?? throw new ArgumentNullException(nameof(groupStore));
+            _logger = logger;
+        }
+
+        public async Task<GroupMigrationResult> MigrateDuplicateGroups()
+        {
+            var groups = (await _groupStore.GetAll()).ToList();
```
michaelvidal24: Should we only get non deleted groups? Might be weird if we merge groups when a groups is deleted...along those lines, will a deleted group contribute to the root cause of the original error?

Fabric.Authorization.Domain/Services/GroupMigratorService.cs
```diff
@@ -0,0 +1,158 @@
...
+                {
+                    DuplicateCount = duplicateGroups.Count
+                };
+
+                foreach (var duplicateGroup in duplicateGroups)
+                {
+                    // migrate roles
+                    foreach (var role in duplicateGroup.Roles)
+                    {
+                        var roleExists = originalGroup.Roles.Any(r => r.Equals(role));
```
michaelvidal24: Should we only migrate non deleted Roles?

Fabric.Authorization.Domain/Services/GroupMigratorService.cs
```diff
@@ -0,0 +1,158 @@
...
+                                _logger.Error(e, msg);
+                                groupMigrationRecord.Errors.Add($"{msg} ({e.Message})");
+                            }
+                        }
+                    }
+
+                    // migrate users (applies custom groups only)
+                    foreach (var user in duplicateGroup.Users)
+                    {
+                        var userExists = originalGroup.Users.Any(u => new UserComparer().Equals(u, user));
```
michaelvidal24: Should we only migrate non deleted users?

Fabric.Authorization.Domain/Services/GroupMigratorService.cs
bsmithvt: I'll update `GetAll` to filter out all deleted child entities.

PR #396, opened 2018-11-30
User/bsmithvt/access control error messages

bsmithvt:


bsmithvt:
I'm thinking I can remove lines 228 and 275 in `member.component.ts` and line 447 in `custom-group.component.ts` to prevent duplicate toasters from displaying.

Fabric.Authorization.AccessControl/src/app/services/global/alert.service.ts
```diff
@@ -8,11 +8,14 @@ export class AlertService {
 
   constructor(private toastr: ToastrService) { }
 
-  public showSyncWarning(errorMessage: string){
-    this.toastr.warning("Changes have been saved, however we were not able to sync the changes to EDW Console. If the user(s) need admin access to EDW Console contact your administrator to manually grant access. The error is: " + errorMessage);
+  public showSyncWarning(errorMessage: string) {
```
ctediz: How do we differentiate between a sync error and a normal error? I'm thinking the interceptor will catch the same error when we're trying to show a sync warning as well.

bsmithvt: @ctediz, that's a good catch, but I think we're ok. The interceptor is doing the same thing as it does today except showing alerts. The sync module only returns 404s as part of the normal logic. We ignore that when showing the general alert in the interceptor, but we still rethrow the error from the interceptor and catch it in the calling code, which hasn't changed and still invokes the sync error alert where appropriate. Worst case is if the sync API returns a 500, we will probably show 2 alerts - 1 general alert in the interceptor and then another sync-specific alert in the calling code where the exception is caught.

bsmithvt: This is certainly something we will want to test though.

Fabric.Authorization.AccessControl/src/app/services/interceptors/fabric-http-error-handler-interceptor.service.ts
```diff
@@ -33,6 +37,8 @@ export class FabricHttpErrorHandlerInterceptorService {
               response.error
             )}`
           );
+
+          this.showAlert(response, response.statusText);
```
michaelvidal24: Is the response.statusText informative enough? Not sure what the alternative would be though.

Fabric.Authorization.AccessControl/src/app/services/global/alert.service.ts
michaelvidal24: Can we setup an automated test for this? This seems like something we should be able to test sufficiently by mocking and stubbing things out...

PR #397, opened 2018-12-03
User/bsmithvt/hot fix merge

bsmithvt:


Fabric.Authorization.Persistence.SqlServer/Stores/SqlServerGroupStore.cs
```diff
@@ -303,6 +303,8 @@ public async Task<Group> AddRolesToGroup(Group group, IEnumerable<Role> rolesToA
         {
             var groupEntity = await AuthorizationDbContext.Groups.SingleOrDefaultAsync(g =>
                 g.Name == group.Name
+                && g.TenantId == group.TenantId
```
michaelvidal24: Do we have to worry about these fields being null for existing groups?

bsmithvt: Yes, they can be null, but this is the same `Where` clause that's used in other methods, and in the case those are null, it will just match on name. We have tests around those scenarios.

PR #400, opened 2018-12-06
User/bsmithvt/custom group null idp support

bsmithvt:


Fabric.Authorization.API/RemoteServices/IdentityProviderSearch/Providers/IdPSearchProvider.cs
```diff
@@ -23,7 +23,7 @@ public class IdPSearchProvider : IIdPSearchProvider
         private readonly ILogger _logger;
 
         private const string ServiceName = "Fabric.IdentityProviderSearchService";
-        private const string PrincipalsEndpoint = "v1/principals/";
+        private const string PrincipalsEndpoint = "principals/";
```
michaelvidal24: Is this the fix for the bug regarding adding an Azure AD group to a role?

bsmithvt: @michaelvidal24, yes. I'll move the bug over once this is merged.

bsmithvt: @michaelvidal24 yes

PR #401, opened 2018-12-07
fixed functional test failures related to new group validations and I…

bsmithvt:
…dentityProvider field


Fabric.Authorization.FunctionalTests/tests/authorizationtests.js
```diff
@@ -56,7 +56,8 @@ describe("authorization tests", function () {
 
     var groupNonCustom = {
         "groupName": "FABRIC\\\Health Catalyst Non-Custom",
-        "groupSource": "Active Directory"
+        "groupSource": "Active Directory",
+		"identityProvider": "Windows"
```
michaelvidal24: Do these changes signal that we made breaking changes to the API?

PR #421, opened 2018-12-27
Added Azure AD user name parsing for the navbar

hc-jordenlowe:
Considering each identity provider will have a different profile object, I split the parsing off for Azure AD.  I more robust solution can come up in the future.

The parsing is a bit tedious, but the code is done.

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/navbar/navbar.component.ts
+++ b/Fabric.Authorization.AccessControl/src/app/navbar/navbar.component.ts
@@ -34,10 +34,62 @@ export class NavbarComponent implements OnInit {
     if (user && user.profile) {
       if (user.profile.family_name && user.profile.given_name) {
         return (user.profile.given_name + ' ' + user.profile.family_name);
-      }else {
+      } else if (user.profile.idp === 'AzureActiveDirectory') {
+        return this.parseAzureADProfileName(user.profile);
+      } else {
         return user.profile.name;
       }
     }
     return '';
+  }
+
+  /*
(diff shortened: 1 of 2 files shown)


hc-jordenlowe:
I will be adding unit tests now to tackle the different scenarios.

Fabric.Authorization.AccessControl/src/app/navbar/navbar.component.ts
```diff
@@ -34,10 +34,55 @@ export class NavbarComponent implements OnInit {
...
+    Azure AD. It will look to see if it is an array and decided if it is able
+    to find a non-email name.  If it cant find that non-email name then it
+    will use the User Principal name,  i.e. email@domain
+  Parameters: profile, the profile of the user from OIDC
+  Returns: string, the name to use in the nav bar for Azure AD
+  */
+  parseAzureADProfileName(profile: any): string {
+    // 1) if the profile does not have an array,
+    // then return the name completely
+    if (!Array.isArray(profile.name)) {
```
ryanorbaker: not for sure on TS, but f you try to access var.Prop (as array) and prop is not present, it should just skip it right?  and then nothing will be returned.

hc-jordenlowe: I see, so I need to make sure name exists too.

hc-jordenlowe: i have added an if above saying, if no name property exists, or if it is null, return ''no name detected''

PR #422, opened 2018-12-27
Only sync users in DosAdmins group

bskeen:


Fabric.Authorization.UnitTests/EDWAdminSyncService/EDWAdminRoleSyncServiceTests.cs
```diff
@@ -1,14 +1,14 @@
...
-using Fabric.Authorization.Domain.Models;
-using Fabric.Authorization.Domain.Services;
-using Fabric.Authorization.Domain.Stores;
-using Moq;
-using Xunit;
-
-namespace Fabric.Authorization.UnitTests.EDWAdminSyncService
+﻿namespace Fabric.Authorization.UnitTests.EDWAdminSyncService
 {
+    using System;
```
hc-jordenlowe: while I would normal suggest this change, if you look at the rest of the code in this solution, you will see the style does not support this change.

To keep consistent, i would change it back.

Fabric.Authorization.UnitTests/EDWAdminSyncService/EDWAdminRoleSyncServiceTests.cs
```diff
@@ -335,34 +355,34 @@ public async Task SyncPermissions_NullUserArrayDoesNotThrowsExceptionAsync()
...
-                        Groups = new List<Group>() { DeletedGroup }
+                        Roles = new Role[] { NoAdminRole },
+                        Groups = new List<Group>() { DosAdminsDeletedGroup }
                     }
                 },
                 0,
                 2
             },
-            new object[] // If a group is deleted, then remove all non admins
+            new object[] // If DosAdmins group is deleted, then remove all users (including admins)
```
hc-jordenlowe: Just to follow the same pattern that the previous test was trying to convey, i would add a group/role that is removed and also one that is added, that way it shows that 1 and 1 scenario.

PR #423, opened 2019-01-04
added validation to ensure valid group source

bsmithvt:


Fabric.Authorization.Domain/Validators/GroupValidator.cs
```diff
@@ -25,6 +25,10 @@ private void ConfigureRules()
                 .WithMessage("Please specify a Source for this Group.")
                 .WithState(g => ValidationEnums.ValidationState.MissingRequiredField);
 
+            RuleFor(group => group)
```
michaelvidal24: Do we have logic that defaults the groupSource anywhere for backward compatibility purposes?

bsmithvt: @michaelvidal24 yes. We were always defaulting it if it is empty prior to running the validation logic.

PR #428, opened 2019-01-10
Code to fix duplicates bug

hckenmiller:


Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -515,9 +515,19 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
       takeUntil(this.ngUnsubscribe))
       .subscribe(null, (error) => {
         if (error.statusCode === 409) {
-          this.groupNameInvalid = true;
-          this.groupNameError = `Could not create group named "${this.groupName}". ` +
-          `A group with the same name exists as a Custom group or a Directory group`;
+          if (error.message.startsWith("The associated user or group name should not be the same as an existing custom group:"))
```
bsmithvt: Can we just use the `groupNameInvalid` and `groupNameError` fields for both cases and get rid of the if/else? What is `error.message` set to in the `else` case where we're hard-coding our own error message? Can we just always take the `error.message` that comes back in the API response?

hckenmiller: It shows up in the wrong spot, need to use the one I created for the initial weirdness bug.

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -215,6 +224,15 @@ public async Task<Group> AddChildGroups(Group group, IEnumerable<Group> childGro
             // first make sure all the child groups are created
             var childGroupList = childGroups.ToList();
 
+            var customGroupChildGroupWithSameName = (await _groupStore.Get(
+                                childGroupList.Select(g => new GroupIdentifier { GroupName = g.GroupIdentifier.GroupName }), true)).ToList();
+
+            if (customGroupChildGroupWithSameName.Any())
+            {
+                throw new AlreadyExistsException<Group>(
+                   $"The associated user or group name should not be the same as an existing custom group: {string.Join(", ", customGroupChildGroupWithSameName.Select(g => g.Name))}");
```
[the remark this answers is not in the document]
hckenmiller: we made users part of this, based on the initial weirdness bug.

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -151,6 +151,15 @@ public async Task<Group> DeleteRolesFromGroup(GroupIdentifier groupIdentifier, I
         public async Task<Group> AddUsersToGroup(GroupIdentifier groupIdentifier, IList<User> usersToAdd)
         {
             var group = await _groupStore.Get(groupIdentifier);
+
```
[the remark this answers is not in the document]
hckenmiller: we made users part of this, based on the initial weirdness bug.

ryanorbaker:
based on my limited exposure to this code it makes sense.   as i understand it the changes prevent the posting of a group that mimics the name of an existing user.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
bsmithvt: I'm just more concerned the `if` statement uses message text returned from the API, so if that changes, then this has to change.

Fabric.Authorization.Domain/Services/GroupService.cs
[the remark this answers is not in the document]
bsmithvt: Is having a user with the same name as a group an issue since those are accessed via the `/user` endpoint and are backed by another table?

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
hckenmiller: Should I throw a different error instead of 409, so that text won't be an issue?

bsmithvt: Not sure we should change the HTTP response if 409 is the appropriate one for this scenario...

PR #433, opened 2019-01-17
fixed bug where missing permission was duplicated in tooltip message

bsmithvt:


Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -134,14 +134,13 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
         ).pipe(
         switchMap((roles: IRole[]) => {
           this.roles = roles;
-
           return this.currentUserService.getPermissions();
         }),
         tap(p => {
```
ctediz: doesn't look like the p (permissions from userService?) are being used here

bsmithvt: @ctediz good catch - it was adding the missing permission once for each role attached to the group. Let me re-check this to ensure this is ok. I may need to check that it's not included in the permissions array and it's not already in the `missingPermissions` array.

PR #447, opened 2019-02-01
User/jason/functionaltests

ctediz:
Add/remove role functional tests for access control


Fabric.Authorization.AccessControl/e2e/pages/access-control.page.ts
```diff
@@ -0,0 +1,60 @@
+import { browser, protractor, element, by } from 'protractor';
+import { tellNgZoneItHasNoPendingMacroTasks } from '../hacks';
+
+export class AccessControlPage {
```
michaelvidal24: We don't have to do this now, but perhaps we consider breaking this page object model up a little bit so we have a page object for each screen in the UI? E.g. the main landing page, the directory user/group page, and the custom group page? This might make the tests more maintainable going forward.

PR #455, opened 2019-02-08
Run protractor tests through npm

ctediz:


Fabric.Authorization.AccessControl/angular.json
```diff
@@ -152,8 +152,7 @@
         "e2e": {
           "builder": "@angular-devkit/build-angular:protractor",
           "options": {
-            "protractorConfig": "./protractor.conf.js",
```
miketabet: Are we guaranteed that the current directory is always going to be the same?  If not, should probably convert this to an absolute path based off the location of this file.  If CD is always the same for protractor, ignore comment.

ctediz: Did some playing around, and that does seem to be the case

PR #456, opened 2019-02-11
Azure AD User migration

hckenmiller:


Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
```diff
@@ -0,0 +1,390 @@
...
+        $connection.Open()    	
+        $result = $command.ExecuteReader()	
+        $userTable = New-Object System.Data.DataTable # Create a DataTable
+        $userTable.Load($result) # Load the results from the DataReader into the DataTable
+        $userTable.Columns['sourceAnchor'].ReadOnly = $false
+        $userTable.Columns['objectId'].ReadOnly = $false
+        $connection.Close()        	
+    }	
+    catch [System.Data.SqlClient.SqlException] {
+        Write-DosMessage -Level "Fatal" -Message "An error ocurred while executing the command. Please ensure the connection string is correct and the authorization database has been setup. Connection String: $($connectionString). Error $($_.Exception.Message)"
```
bsmithvt: Do we not want to print out `$_.Exception` vs. `$_.Exception.Message`?

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
```diff
@@ -0,0 +1,390 @@
...
+            $command.CommandText = "IF NOT EXISTS 
+                                    (SELECT 1 FROM [dbo].[GroupUsers] g
+                                     WHERE SubjectId = @subjectId
+                                     AND IdentityProvider = 'OpenIDConnect:CatalystAzureAD')
+                                    BEGIN
+                                    INSERT INTO [dbo].[GroupUsers] ([SubjectId], [IdentityProvider], [CreatedBy], [CreatedDateTimeUtc], [GroupId], [IsDeleted])
+                                    VALUES(@subjectId, @identityProvider, @createdBy, @createdDateTime, @groupId, @isDeleted)
+                                    END;"
+            $command.Parameters["@subjectId"].Value = $user.objectId
+            $command.Parameters["@identityProvider"].Value = "OpenIDConnect:CatalystAzureAD"
```
bsmithvt: Same question as above.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
```diff
@@ -0,0 +1,390 @@
...
+            FROM [dbo].[Users] u	
+            LEFT OUTER JOIN [dbo].[GroupUsers] g	
+                ON u.SubjectId = g.SubjectId AND u.IdentityProvider = g.IdentityProvider
+		    LEFT OUTER JOIN [dbo].[RoleUsers] r
+                ON u.SubjectId = r.SubjectId AND u.IdentityProvider = r.IdentityProvider
+            LEFT OUTER JOIN [dbo].[UserPermissions] p	
+                ON u.SubjectId = p.SubjectId AND u.IdentityProvider = p.IdentityProvider
+            WHERE u.IsDeleted = 0
+				AND u.IdentityProvider = 'Windows'
+				AND u.IdentityProviderUserPrincipalName IS NOT NULL";
```
bsmithvt: Is this guaranteed to be populated when this script is run?

bsmithvt:
In general, do we want to wrap any SQL commands in `try/catch` blocks and log errors. Also, I would add more info logs stating how many records were found to migrate, how many records were migrated, etc. for troubleshooting purposes.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
hckenmiller: If it meets the users meet the where constraint.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
```diff
@@ -0,0 +1,390 @@
...
+                                   WHERE SubjectId = @subjectId 
+                                   AND IdentityProvider = 'OpenIDConnect:CatalystAzureAD')
+                                  BEGIN
+                                  INSERT INTO [dbo].[Users] ([SubjectId], [ParentUserId], [IdentityProvider], [CreatedBy], [CreatedDateTimeUtc], [IsDeleted])
+                                  VALUES(@subjectId, @parentId, @identityProvider, @createdBy, @createdDateTime, @isDeleted)
+                                  END;"
+          $command.Parameters["@subjectId"].Value = $user.objectId
+          $command.Parameters.Add("@parentId", [System.Data.SqlDbType]::Int)
+          $command.Parameters["@parentId"].Value = $user.Id
+          $command.Parameters["@identityProvider"].Value = "OpenIDConnect:CatalystAzureAD"
```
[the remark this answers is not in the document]
hckenmiller: I am still trying to figure out where the other value comes into play. The bug was marked done by Mike V. so I figure it is happening somewhere.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
hckenmiller: Exception.ToString() would give us more verbose output, including a stack trace, which would be more valuable for debugging. I was just following what we are doing currently.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
bsmithvt: I was asking if `IdentityProviderUserPrincipalName` guaranteed to be `NOT NULL`?

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
bsmithvt: I think we're doing `$_.Exception` in some places but using just `.Message` in other places.

bsmithvt: I don't think you should need the `ToString()`.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
hckenmiller: I thought that this value was getting populated, but it doesn't look like Name or IdentityProviderUserPrincipalName are always there. For instance, when we install Auth and it adds the user, the Name and IdentityProviderUserPrincipalName aren't added, but if a user is added in Access Control, I believe it is added to IdentityProviderUserPrincipalName each time. So I may have to use SubjectId instead. What are your findings?

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
hckenmiller: Maybe not in powershell, that is just the documentation I have come across in .Net framework.

Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
bsmithvt: I didn't think it was always non-null, so I think you'll need to use another field.

hckenmiller: I will just use SubjectId since it has to have a value. IdentityProviderUserPrincipalName was a column added later, and we avoided having to do a migration script by adding it as a nullable. I think the intention though was to always populate it.

PR #458, opened 2019-02-15
fixed some typo/logging issues

bsmithvt:


Fabric.Authorization.API/scripts/AzureAD-Groups-Migration.ps1
```diff
@@ -217,7 +217,6 @@ function Move-ActiveDirectoryGroupsToAzureAD {
                     Write-DosMessage -Level Information -Message "Found group $($azureADGroup.DisplayName) with ObjectId $($azureADGroup.ObjectId) in Azure AD Tenant $($tenantId)"
                 }
                 else {
```
ctediz: this else can probably removed right?

Fabric.Authorization.API/scripts/AzureAD-Utilities.psm1
```diff
@@ -24,13 +24,13 @@ if ($azureAD.length -eq 0) {
     $installed = Get-InstalledModule -Name AzureAD -ErrorAction "silentlycontinue"
 
     if (($null -eq $installed) -or ($installed.Version.CompareTo($minVersion) -lt 0)) {
-        Write-Host "Instgalling AzureAD from Powershell Gallery"
+        Write-DosMessage "Installing AzureAD from Powershell Gallery"
```
michaelvidal24: Does Write-DosMessage require the -Level parameter?

Fabric.Authorization.API/scripts/AzureAD-Groups-Migration.ps1
bsmithvt: It prevents the second `if` check below. So I suppose we could remove that check or this `else`.

ctediz: Ah, I assumed the catch was the end of the foreach

PR #460, opened 2019-02-20
Added code to use identity url instead of discovery

hc-jordenlowe:
This is the first PR to test the waters on this approach.

First, if discovery service root is populated, it will revert to the old way.  I will be adding the feature flag to authorization service because that seems to be the better way to do it.  If I should add the feature flag explicitly in Access Control, let me know.

Second, because we got Access Control's url from discovery and we need it to get an access token, I am saying authorization must push both Identity and Access Control's url.

If everything looks good i will fix up the code some more, add in authorization code changes and add tests.

### DIFF
--- a/Fabric.Authorization.API/Configuration/DiscoveryServiceSettings.cs
+++ b/Fabric.Authorization.API/Configuration/DiscoveryServiceSettings.cs
@@ -2,9 +2,11 @@
 {
     public class DiscoveryServiceSettings
     {
-        public string AccessControlToken { get; set; }
-        public bool UseDiscovery { get; set; }
-        public string Endpoint { get; set; }
+        public string DiscoveryServiceToken { get; set; }
         public string IdentityServiceToken { get; set; }
+        public string AccessControlUIToken { get; set; }
+        public bool UseDiscovery { get; set; }
+        public bool UseOAuth2Authentication { get; set; }
+        public string Endpoint { get; set; }
     }
 }
--- a/Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
+++ b/Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
@@ -1,4 +1,4 @@
-import { map, mergeMap } from 'rxjs/operators';
+import { map, mergeMap, switchMap } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { HttpClient } from '@angular/common/http';
 import { Observable, of } from 'rxjs';
@@ -47,14 +47,7 @@ export class ServicesService {
     constructor(private http: HttpClient, private configService: ConfigService) {}
 
     public initialize(): Observable<string> {
-        var observable: Observable<string> = null;
-        if(this.configService.getIdentityServiceRoot()) {
-            observable = of(this.discoveryServiceEndpoint)
-        } else {
-            observable = this.configService.getDiscoveryServiceRoot()
-        }
-
-        return observable.pipe(
+        return this.discoveryServiceEndpoint.pipe(
             mergeMap(discoveryServiceRoot => {
                 const url: string =
                     `${discoveryServiceRoot}/Services?$filter=` +
@@ -69,7 +62,7 @@ export class ServicesService {
                     );
 
                     if (service.name === 'DiscoveryService') {
-                        service.url = this.discoveryServiceEndpoint;
+                        service.url = null;
                     } else if (targetService === undefined) {
                         throw new Error(`The ${service.name} was not found in discovery service. Please ensure it is set up correctly.`);
                     } else {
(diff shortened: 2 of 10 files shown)


Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -56,17 +67,32 @@ export class ServicesService {
                     const targetService: IDiscoveryService = response.value.find(
                         s => s.ServiceName === service.name && (!service.version || s.Version === service.version)
                     );
-                    if (targetService === undefined) {
+
+                    if (service.name === 'DiscoveryService') {
+                        service.url = this.discoveryServiceEndpoint;
```
hc-jordenlowe: Ill have to double check this returns the correct discovery endpoint

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -37,13 +37,24 @@ export class ServicesService {
...
+            requireAuthToken: true
         }
     ];
 
     constructor(private http: HttpClient, private configService: ConfigService) {}
 
     public initialize(): Observable<string> {
-        return this.configService.getDiscoveryServiceRoot().pipe(
+        var observable: Observable<string> = null;
+        if(this.configService.getIdentityServiceRoot()) {
```
hc-jordenlowe: I think i have to flip these

Fabric.Authorization.AccessControl/src/index.html
```diff
@@ -9,6 +9,9 @@
   <link rel="icon" type="image/x-icon" href="favicon.ico">
   <script>
     window.discoveryServiceRoot = "DISCOVERY_SERVICE_ROOT";
+
+    window.identityServiceRoot = "IDENTITY_SERVICE_ROOT";
```
hc-jordenlowe: might have to change the authorization logic to insert empty strings and then look for empty strings instead

Fabric.Authorization.API/Configuration/DiscoveryServiceSettings.cs
```diff
@@ -3,7 +3,10 @@
     public class DiscoveryServiceSettings
     {
         public string AccessControlToken { get; set; }
```
michaelvidal24: Does it make sense to rename this to DiscoveryServiceToken?

Fabric.Authorization.API/Configuration/DiscoveryServiceSettings.cs
```diff
@@ -3,7 +3,10 @@
...
+        public string IdentityServiceToken { get; set; }
+        public string AccessControlUIToken { get; set; }
```
michaelvidal24: Do we need this because we need to setup the reply URL?

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -56,17 +67,32 @@ export class ServicesService {
...
-
-                return this.services.find(s => s.name === 'IdentityService').url;
             })
         );
     }
 
+    private baseDiscoveryServiceUrl : string;
+    get discoveryServiceEndpoint(): string {
+        if(this.baseDiscoveryServiceUrl === undefined) {
+            const identityRootUrl: string = this.configService.getIdentityServiceRoot();
```
michaelvidal24: In the oidc client, there is a metadataService object that we might consider leveraging: https://github.com/IdentityModel/oidc-client-js/blob/dev/src/MetadataService.js.

Fabric.Authorization.API/Configuration/DiscoveryServiceSettings.cs
hc-jordenlowe: Its used to generate the token with Access control.  The UI makes a call back to OICD.html, which need the current location.  I could get the current location but thought that if we set this somewhere like auth service, then it would be more consistant.

i.e. avoids localhost/<hostname>/127.0.0.1 domain name confusion

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
hc-jordenlowe: ill check this out

Fabric.Authorization.API/Configuration/DiscoveryServiceSettings.cs
hc-jordenlowe: sounds good

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
hc-jordenlowe: I see this js object will parse the metadata of the document into a consumable object.  Yeah i can see fitting this in

hc-jordenlowe:
Closed due to duplicate

PR #462, opened 2019-02-22
Display error page when cookies are not enabled

rlipenta-hc:
This PR pertains to bug [162684](https://healthcatalyst.visualstudio.com/CAP/_workitems/edit/162684). 

Within `main.ts` use the new `BrowserRequirementsService`'s `cookiesEnabled` method to test whether or not cookies are enabled. If they are not bootstrap the new `no-cookies.app` instead of the main application. The main application itself requires cookies along with local and session storage (as controlled by browsers cookies settings). By creating the `no-cookies.app` we can provide the user with an appropriately styled error page. The error page will retest for cookies and reload the application when cookie support is re-enabled.

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/services/browser-requirements.service.ts
+++ b/Fabric.Authorization.AccessControl/src/app/services/browser-requirements.service.ts
@@ -1,0 +1,26 @@
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class BrowserRequirementsService {
+
+  constructor() { }
+
+  cookiesEnabled(): boolean {
+    const userAgent = window.navigator.userAgent;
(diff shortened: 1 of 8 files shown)


Fabric.Authorization.AccessControl/src/app/services/browser-requirements.service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  constructor() { }
+
+  cookiesEnabled(): boolean {
+    const userAgent = window.navigator.userAgent;
+    if (!(userAgent.indexOf('MSIE ') > -1 || userAgent.indexOf('Trident') > -1) &&
+        typeof navigator.cookieEnabled !== 'undefined' && navigator.cookieEnabled) {
+      return true;
+    }
+    if (!document.cookie || document.cookie === '') {
```
hc-jordenlowe: looks like this tests both chrome based browsers and firefox, is that right?

rlipenta-hc: It tests appropriately for Chrome, Firefox and IE.

PR #463, opened 2019-02-25
Updates to make it easier to debug the angular application

michaelvidal24:
Changes include:

- Introducing an http interceptor so we can mock the response from discovery service that is wired up in DI only when we are not running in production mode.
- New debug profile to run Fabric.Authorization through IIS
- Added a new `npm run watch` script to register access control with the installed version of identity


Fabric.Authorization.API/Infrastructure/Middleware/AngularMiddleware.cs
```diff
@@ -37,12 +37,16 @@ public async Task Invoke(HttpContext context)
                 var fullPath = Path.Combine(_hostingEnvironment.WebRootPath, AccessControl.Path, AccessControl.Index);
                 _indexContent = File.ReadAllText(fullPath);
 
-                var discoveryServiceSettings = _appConfiguration.DiscoveryServiceSettings;
+                var discoveryServiceSettings = _appConfiguration.DiscoveryServiceSettings;
+                var identityServerSettings = _appConfiguration.IdentityServerConfidentialClientSettings;
```
hc-jordenlowe: I see you have added some of the things I am working on.  Probably needed for the move.  Ill keep this in mind during merge time.

michaelvidal24: Yup - I tried to use the same pattern you used so it would make the merge easier...

Fabric.Authorization.AccessControl/src/app/services/interceptors/fabric-http-fake-discovery-interceptor.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    if(req.url.includes('DiscoveryService/v1/Services?$filter=ServiceName')){
+        this.configService.getIdentityServiceRoot().subscribe(url => {
+          this.responseBody = {
+            "@odata.context":"http://localhost/DiscoveryService/v1/$metadata#Services(ServiceUrl,Version,ServiceName)","value":[
+              {
+                "ServiceUrl":"http://localhost/IdentityProviderSearchService/v1","Version":1,"ServiceName":"IdentityProviderSearchService"
+              },{
+                "ServiceUrl":url,"Version":1,"ServiceName":"IdentityService"
+              },{
+                "ServiceUrl":"http://localhost/AuthorizationDev/v1","Version":1,"ServiceName":"AuthorizationService"
```
hc-jordenlowe: Is the localhost going to cause problems if everything else is registered with domain name?  Maybe all you need from discovery service is IdPSS

Fabric.Authorization.AccessControl/src/app/services/interceptors/index.ts
```diff
@@ -2,17 +2,25 @@
...
-  },
-  {
-    provide: HTTP_INTERCEPTORS,
-    useClass: FabricHttpErrorHandlerInterceptorService,
-    multi: true
-  }
-];
+  });
+}
+httpInterceptorProviders.push({
```
bsmithvt: Should this block be wrapped in an `else`?

Fabric.Authorization.AccessControl/src/app/services/interceptors/fabric-http-fake-discovery-interceptor.service.ts
michaelvidal24: The key here is that we are using a fully qualified url for interacting with identity, which will come from appsettings.json. Everything else should work correctly from then on.

Fabric.Authorization.AccessControl/src/app/services/interceptors/index.ts
michaelvidal24: No - we always want these interceptors added, we just want to conditionally add the fake interceptor to allow mocking only when we don't build in production mode.

PR #464, opened 2019-02-26
script changes

hckenmiller:
script changes to fix bugs

### DIFF
--- a/Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
+++ b/Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
@@ -415,7 +415,7 @@ function Get-ADUsers
          $userMap = $entry."ms-Ds-ConsistencyGuid" 
          if (![string]::IsNullOrEmpty($userMap))
          {
-           # put the user ms-DS-ConsistenceyGuid into base 64 string format to compare against the AzureAD ImmutableId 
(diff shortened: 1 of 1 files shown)


Fabric.Authorization.API/scripts/AzureADUsers-Migration.ps1
```diff
@@ -154,130 +220,158 @@ function Add-AuthUsers
...
                 {   
-			      $totalGroupUsersAdded += $resultGroupUsers
+			        $totalGroupUsersAdded += $resultGroupUsers
 	    	    }
 
 			    if($groupCount -eq 1)
 			    {
-			      $command.Parameters.RemoveAt("@groupId")
-				  $groupCount = 0
+			        $command.Parameters.RemoveAt("@groupId")
```
michaelvidal24: It's tough to tell from the formatting, but is this inside the foreach loop? Are we using it as a way to clear out that parameter on each iteration of the loop so we can update the groupId param for each iteration?

hckenmiller: I actually could move that out of the foreach loop, before the next foreach loop for RoleUsers. Parameterized queries don't like unused parameters, and an error is thrown if there are any unused. Since GroupUsers and RoleUsers have at least one different column, I have to remove it before executing the next query.

PR #466, opened 2019-03-01
Changed toast error from displaying subjectId to new claim called name

joshnielsen10:
Added a new Claim field called Name, created a new property in FabricModule called Name and retrieved the claim called Name for the current user, and passed that property through the methods instead of subjectId.

### DIFF
--- a/Fabric.Authorization.API/appsettings.json
+++ b/Fabric.Authorization.API/appsettings.json
@@ -11,7 +11,7 @@
     "Password": ""
   },
   "IdentityServerConfidentialClientSettings": {
-    "Authority": "https://HC2822.hqcatalyst.local/identity",
+    "Authority": "http://localhost/identity",
     "ClientId": "authorization-api",
     "ClientSecret": "secret",
     "Scopes": [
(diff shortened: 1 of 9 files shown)


Fabric.Authorization.API/appsettings.json
```diff
@@ -11,7 +11,7 @@
     "Password": ""
   },
   "IdentityServerConfidentialClientSettings": {
-    "Authority": "http://localhost/identity",
+    "Authority": "https://HC2822.hqcatalyst.local/identity",
```
bsmithvt: Let's revert this back to `http://localhost`

Fabric.Authorization.API/web.config
```diff
@@ -26,7 +25,7 @@
       <remove name="WebDAV" />
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="dotnet" arguments=".\Fabric.Authorization.API.dll" stdoutLogEnabled="true" stdoutLogFile=".\logs\stdout" >
```
ctediz: this seems like a weird change, was that intentional?

joshnielsen10: No it wasn't, me and Brian were talking about it. It happened on mine and Jorden's using the new Access Control debug process. I'm removing it and creating new commits so it will be what it originally was. But it's something we should probably talk to Mike about.

PR #467, opened 2019-03-05
Fabric Authorization uses Identity for discovery document

hc-jordenlowe:


Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -37,73 +37,167 @@ export class ServicesService {
...
+    get isOAuthAuthenticationEnabled(): Observable<boolean> {
+        return this.configService.getUseOAuthAuthentication().pipe(
+            switchMap(stringBool => stringBool === 'true' ? of(true) : of(false))
+        )
+    }
+
+    get discoveryServiceUrlFromIdentity(): Observable<string> {
+        return this.configService.getIdentityServiceRoot()
+            .pipe(map((url) => `${url}/.well-known/openid-configuration`),
+                switchMap((url) => this.http.get(url)),
```
michaelvidal24: if this call fails, how do we handle the error?

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -37,73 +37,167 @@ export class ServicesService {
...
-       services.forEach(element => {
-        var result = url.toLowerCase().includes(element.url.toLowerCase());
-        if (result)
-        {
-          if (element.url.length > serviceUrlMatch.length)
-          {
-            serviceUrlMatch = element.url.toLowerCase();
-          }
+    private parseUrl(url) {
+        var location = document.createElement("a")
```
michaelvidal24: Is there a simpler way to parse a url: https://developer.mozilla.org/en-US/docs/Web/API/URL

Fabric.Authorization.API/web.config
```diff
@@ -1,38 +1,38 @@
...
+            <add input="{UNENCODED_URL}" pattern="(.*)(%5C)(.*)" />
+          </conditions>
+        </rule>
+      </rules>
+    </rewrite>
+    <handlers>
+      <remove name="WebDAV" />
+      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
+    </handlers>
+    <aspNetCore processPath="bin\IISSupport\VSIISExeLauncher.exe" arguments="-argFile IISExeLauncherArgs.txt" stdoutLogEnabled="true" stdoutLogFile=".\logs\stdout">
```
bsmithvt: I think this changed because of Mike's dev environment changes. Josh noticed this in his PR and we reverted it. Do we want this change permanently?

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -37,73 +37,167 @@ export class ServicesService {
...
+        // to populate these additional fields.
+        if (location.host == "") {
+            location.href = location.href;
         }
-       });
-       var matchedService = services.find(s => s.url.toLowerCase() === serviceUrlMatch);
-       const serviceItem: IService = matchedService;
-       return serviceItem;
+
+        return location
```
bsmithvt: Missing semicolon.

Fabric.Authorization.AccessControl/src/app/services/fabric-base.service.ts
```diff
@@ -8,5 +8,7 @@ export class FabricBaseService {
   constructor(
     protected httpClient: HttpClient,
     @Inject('IAccessControlConfigService') protected accessControlConfigService: IAccessControlConfigService
-  ) {}
+  ) {
+    accessControlConfigService.getBaseUrls().toPromise().then(f => console.log)
```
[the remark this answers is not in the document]
hc-jordenlowe: I could not get this to evaluate the observable until i created a promise.  Ill see if i can remove the then

Fabric.Authorization.API/web.config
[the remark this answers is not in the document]
hc-jordenlowe: reverted

hc-jordenlowe: reverted

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -37,73 +37,167 @@ export class ServicesService {
...
+            switchMap(isEnabled => {
+              if(isEnabled) {
+                // if OAuth, we get identity and access control up front
+                return forkJoin(this.identityServiceEndpoint, this.accessControlEndpoint)
+              } else {
+                // If Windows Auth, we have discovery up front and have to build
+                // out the service urls from discovery
+                return this.buildServiceMaps().toPromise().then(identityUrl => {
+                        return this.accessControlEndpoint.toPromise().then(accessControlUrl => {
+                            return [identityUrl, accessControlUrl]
```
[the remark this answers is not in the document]
hc-jordenlowe: makes sense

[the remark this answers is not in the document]
hc-jordenlowe: Now I remember why I did this.  It is a complication of the forkJoin that the results get put into 2 items of an array.  Ill add another line that will normalize this data

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
hc-jordenlowe: Not supported in IE 11.  This was the most browser friendly version.

hc-jordenlowe: using URL would have been way easier

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
hc-jordenlowe: Like our code before, there is no retry logic.  While I dont use this as an excuse, instead we can build an interceptor that would do some retry logic if we want?

I dont know if there is a neat RxJS way to retry, but the interceptor seems like a better idea all around for all calls.

hc-jordenlowe: I see, so we should check the response to make sure its 200, otherwise throw error

hc-jordenlowe: Verified that there is an interceptor that will catch exceptions from the HTTP client and display them in a toast message.  This should be good then

PR #471, opened 2019-03-11
User/jason/duplicategroupnames

ctediz:
Tenant alias appears in idpss search results
Tenant alias appears during member/custom group selections

Going to create tests in another pr

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.ts
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.ts
@@ -166,15 +166,17 @@ export class MemberComponent implements OnInit, OnDestroy {
     this.ngUnsubscribe.complete();
   }
 
+  setSearchText(principal: IFabricPrincipal): void {
+    this.searchText = principal.identityProviderUserPrincipalName || principal.subjectId;
+    if (!!principal.tenantAlias) {
+      this.searchText += '@' + principal.tenantAlias;
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


Fabric.Authorization.AccessControl/src/app/modules/access-control/member/member.component.ts
```diff
@@ -169,7 +169,12 @@ export class MemberComponent implements OnInit, OnDestroy {
   selectPrincipal(principal: IFabricPrincipal): Subscription {
     this.principals = [];
     this.selectedPrincipal = principal;
-    this.searchText = principal.identityProviderUserPrincipalName || principal.subjectId;
+    const alias: string = principal.tenantAlias;
+    if (!alias) {
+      this.searchText = (principal.identityProviderUserPrincipalName || principal.subjectId);
```
hc-jordenlowe: Looks like you could pull this out of the if and do something like this:
this.searchText = (principal.identityProviderUserPrincipalName || principal.subjectId)
if(alias) {
   this.searchText += '@' + alias;
}

that would help reduce the complexity.  I dont know if that or would work the same, might want to test it

ctediz: moved into different function, this is what I came up with:

```
this.searchText = principal.identityProviderUserPrincipalName || principal.subjectId;
    if (!!principal.tenantAlias) {
      this.searchText += '@' + principal.tenantAlias;
    }
```

as long as tenantAlias has a value, append to searchText

PR #473, opened 2019-03-12
Move getPrincipalNameToDisplay logic into service

ctediz:
member-list and custom-group components to use the name service


Fabric.Authorization.AccessControl/src/app/services/name-display.service.ts
```diff
@@ -0,0 +1,20 @@
...
+import { IFabricPrincipal } from '../models/fabricPrincipal.model';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class NameDisplayService {
+  constructor() { }
+
+  getPrincipalNameToDisplay(principal: IFabricPrincipal): string {
+    if (principal.principalType.toLowerCase() === 'user') {
```
bsmithvt: I know we use `user` and `group` constants somewhere else to represent `type`. Perhaps we factor those out into a separate constants file/class?

PR #474, opened 2019-03-14
Build out services from discovery

ctediz:


Fabric.Authorization.AccessControl/src/app/services/global/auth.service.ts
```diff
@@ -64,6 +64,9 @@ export class AuthService implements IAuthService {
 
         this.userManager = new UserManager(clientSettings);
 
+        // build out services now..
+        this.servicesService.initialize();
```
ctediz: Need an auth token to hit discovery, but we need to have a UserManager object created before we can get a token to hit DiscoveryService with an oauth token.
So it's in this weird state were we grab the config values, and then build out the services through buildServiceMaps

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -48,35 +49,34 @@ export class ServicesService {
...
     public getIdentityAndAccessControlUrl(): Observable<UrlResponse> {
         return this.isOAuthAuthenticationEnabled.pipe(
             switchMap(isEnabled => {
-              this.services.find(s => s.name === 'DiscoveryService').requireAuthToken = isEnabled              
-              if(isEnabled) {
+              this.services.find(s => s.name === 'DiscoveryService').requireAuthToken = isEnabled;
+              if (isEnabled) {
                 // if OAuth, we get identity and access control up front
-                return forkJoin(this.identityServiceEndpoint, this.accessControlEndpoint)
+                return forkJoin(this.identityServiceEndpoint, this.accessControlEndpoint);
```
ctediz: Can't build out services early if oauth is enabled for reasons stated above

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -48,35 +49,34 @@ export class ServicesService {
     constructor(private http: HttpClient, private configService: ConfigService) { }
 
     public initialize() {
-        this.buildServiceMaps()
+        this.buildServiceMaps().toPromise();
```
ctediz: Currently Discovery is being called twice on each page load

hc-jordenlowe:
I feel like this should be initialized once, and during the initialization is when it should go out and get the list of service urls from Discovery.  This would require reworking both AuthService and the HttpInterceptor.

Maybe we initialize this closer to the angular application start process. If this class was static or saved somewhere then by the time we get to auth service and the interceptor, all it has to do is call the things it needs.  I just dont know where that would be in Angular.  @michaelvidal24 maybe you got an idea of making this logic more straight forward?

Fabric.Authorization.AccessControl/src/app/services/interceptors/fabric-http-request-interceptor.service.ts
```diff
@@ -38,6 +38,10 @@ export class FabricHttpRequestInterceptorService implements HttpInterceptor {
       );
 
       return tokenObservable.pipe(mergeMap(accessToken => {
+        // do not send request if user token was not found (not authenticated), and was required
+        if (!accessToken) {
+          return empty();
```
ctediz: Prevent trying to send a request to discovery without an auth token and showing a windows auth prompt for a moment.

Caused because the auth.service is trying to populate service.service's service list by calling discovery, which fires off this interceptor, but auth.service has yet to be fully initialized so it can't quite get a user in time to add an access token.

michaelvidal24: Should we also just add a console log here, so we know that we blocked a request?

ctediz: Sure

PR #475, opened 2019-03-15
Fix cookie detection for IE

rlipenta-hc:


hc-jordenlowe:
uh oh, i shouldnt be pressing buttons

PR #478, opened 2019-03-18
User/jason/refactorauthservice

ctediz:


Fabric.Authorization.AccessControl/src/app/services/global/initializer.service.ts
```diff
@@ -1,12 +1,14 @@
...
 @Injectable()
 export class InitializerService {
 
-  constructor(@Inject('IAuthService')private authService: IAuthService) { }
+  constructor(@Inject('IAuthService')private authService: IAuthService, private servicesService: ServicesService) { }
 
   initialize() {
-    return this.authService.initialize();
+    return this.authService.initialize()
+      .then(() => this.servicesService.initialize());
```
ctediz: I believe this is causing the AuthGuard to not fire off because the AuthGuard code isn't created until after this initializer service runs.

Looking to see if there's a better place for creating the services.service's list of services.

Fabric.Authorization.AccessControl/src/app/services/global/auth.service.ts
```diff
@@ -30,14 +30,15 @@ export class AuthService implements IAuthService {
...
+  constructor(private httpClient: HttpClient, private configService: ConfigService) {
     this.clientId = 'fabric-access-control';
   }
 
   initialize(): Promise<any> {
     // Need the initializer to load all the values first
-    // This makes sure we get the correct accessControl endpoint
-    return this.servicesService.getIdentityAndAccessControlUrl().pipe(
+    // This makes sure we get the correct accessControl endpoint, from the configService
+    return forkJoin(this.configService.getIdentityServiceRoot(), this.configService.getAccessControlServiceRoot()).pipe(
```
hc-jordenlowe: this is ok, except when windows auth is enabled.  If windows auth is enabled, then we do not use the config service to get these 2 urls, we use discovery service

Fabric.Authorization.AccessControl/src/app/services/global/services.service.ts
```diff
@@ -49,35 +48,12 @@ export class ServicesService {
...
         return this.isOAuthAuthenticationEnabled.pipe(
             switchMap(isEnabled => {
-              this.services.find(s => s.name === 'DiscoveryService').requireAuthToken = isEnabled;
-              if (isEnabled) {
-                // if OAuth, we get identity and access control up front
-                return forkJoin(this.identityServiceEndpoint, this.accessControlEndpoint);
-              } else {
-                // If Windows Auth, we have discovery up front and have to build
-                // out the service urls from discovery
-                return this.buildServiceMaps().toPromise().then(identityUrl => {
```
hc-jordenlowe: Dont forget about this scenario, if windows auth is enabled, then we have to pull the urls from discovery service

Fabric.Authorization.AccessControl/src/app/services/global/auth.service.ts
ctediz: The thought here was that these two url's are being passed in during authorization install, and we could use those to setup the auth client

hc-jordenlowe: If you look at the service, i think I did make it so Auth pushes all values every time.  Might be safe to do this then.

Have you found a way to get around the multiple discovery service calls?

ctediz: This is what was done initially with the previous pr, which was sending an undefined token because the auth.service wasn't built out yet when the discovery request went off.

https://github.com/HealthCatalyst/Fabric.Authorization/blob/18bdefa0ee645ce52518fefa7c956812f84e3734/Fabric.Authorization.AccessControl/src/app/services/interceptors/fabric-http-request-interceptor.service.ts#L42

Running into a similar issue now with the auth.guard not firing off (currently the services list if being initialized in the initializer service which hooks into the angular initialization).

Fabric.Authorization.AccessControl/src/app/services/global/initializer.service.ts
michaelvidal24: Is there a reason to not simply initialize the services.service in its own constructor, that way it initializes when its needed, instead of eagerly initializing it here? Perhaps with the changes we made to get the identity and access control ui urls from config instead of discovery all the time, that option is now open to us?

PR #482, opened 2019-04-05
Fixing issue where user removes self from group

joshnielsen10:
Fixed the issue when the user removes themselves from the DOS Admin group, if you quickly go back in the group, the permsission cache is still active and you can see an 'Enabled' Save button.

### DIFF
--- a/Fabric.Authorization.API/Dockerfile
+++ b/Fabric.Authorization.API/Dockerfile
@@ -3,7 +3,11 @@ ARG source
 WORKDIR /app
 EXPOSE 5004
 
-RUN apt-get update -y && \
+RUN echo "deb [check-valid-until=no] http://cdn-fastly.deb.debian.org/debian jessie main" > /etc/apt/sources.list.d/jessie.list
+RUN echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
+RUN sed -i '/deb http:\/\/deb.debian.org\/debian jessie-updates main/d' /etc/apt/sources.list
+
+RUN apt-get -o Acquire::Check-Valid-Until=false update -y && \
 	apt-get install -y --no-install-recommends\
 	cron \
 	ldap-utils \
--- a/Fabric.Authorization.API/Dockerfile
+++ b/Fabric.Authorization.API/Dockerfile
@@ -3,23 +3,6 @@ ARG source
 WORKDIR /app
 EXPOSE 5004
 
-RUN echo "deb [check-valid-until=no] http://cdn-fastly.deb.debian.org/debian jessie main" > /etc/apt/sources.list.d/jessie.list
-RUN echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
-RUN sed -i '/deb http:\/\/deb.debian.org\/debian jessie-updates main/d' /etc/apt/sources.list
-
-RUN apt-get -o Acquire::Check-Valid-Until=false update -y && \
-	apt-get install -y --no-install-recommends\
-	cron \
-	ldap-utils \
-	&& rm -rf /var/lib/apt/lists/*
-
 COPY ${source:-obj/Docker/publish} .
-
-RUN mkdir -p /usr/local/bin \
-	&& mv scripts/update-groups.sh /usr/local/bin/update-groups.sh \
-	&& chmod a+x scripts/setup-crontab.sh \
-	&& chmod a+x /usr/local/bin/update-groups.sh
-
-#RUN	./scripts/setup-crontab.sh
 
 ENTRYPOINT ["dotnet", "Fabric.Authorization.API.dll"]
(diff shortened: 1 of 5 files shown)


Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -73,7 +76,9 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
     private alertService: AlertService,
     @Inject('IAccessControlConfigService')
     private configService: IAccessControlConfigService,
-    private userNameService: NameDisplayService
+    private userNameService: NameDisplayService,
+    @Inject('IAuthService')
```
ctediz: does the test file need to be updated to have this added as well

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -593,4 +607,15 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
 
     return user.subjectId;
   }
+
+  checkIfUserRemovedSelfFromGroup(currentUser: User, removedUsersList: IUser[]): boolean {
+    var matched = false;
+    removedUsersList.forEach(removedUser => {
+      if(removedUser.subjectId.toLowerCase == currentUser.profile.sub.toLowerCase) {
```
ctediz: I wonder if we store the user somewhere else. But it looks like you're getting the user direct from the auth service so maybe not

joshnielsen10: I was trying to see if/where the user could be stored and couldn't find it. Doesn't mean we aren't though.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
joshnielsen10: That could be my issue. Running the tests locally as well is giving me a StaticInjectorError[IAuthService]: NullInjectorError: No provider for IAuthService

Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
```diff
@@ -25,4 +25,9 @@ export class CurrentUserService {
 
     return of(this.permissions);
   }
+
+  resetPermssionCache() {
```
hckenmiller: Perm"i"ssion

joshnielsen10: Haha thanks for catching that.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
joshnielsen10: Updated the test file, thanks Jason and Brian for your help with this.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -593,4 +607,15 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
 
     return user.subjectId;
   }
+
+  checkIfUserRemovedSelfFromGroup(currentUser: User, removedUsersList: IUser[]): boolean {
+    var matched = false;
+    for(let i = 0; i < removedUsersList.length; i++){
+      if(removedUsersList[i].subjectId.toLowerCase() == currentUser.profile.sub.toLowerCase()) {
```
bsmithvt: Change `==` to `===`.

Fabric.Authorization.API/Dockerfile
```diff
@@ -3,7 +3,11 @@ ARG source
 WORKDIR /app
 EXPOSE 5004
 
-RUN apt-get update -y && \
+RUN echo "deb [check-valid-until=no] http://cdn-fastly.deb.debian.org/debian jessie main" > /etc/apt/sources.list.d/jessie.list
```
hc-jordenlowe: you should still do the "apt-get update" after these sources are added so that it gets the latest packages.

Fabric.Authorization.API/Dockerfile
```diff
@@ -3,7 +3,11 @@ ARG source
...
+RUN echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
+RUN sed -i '/deb http:\/\/deb.debian.org\/debian jessie-updates main/d' /etc/apt/sources.list
+
+RUN apt-get -o Acquire::Check-Valid-Until=false update -y && \
```
hc-jordenlowe: So, why do we need to modify our docker image to resolve this issue?

Fabric.Authorization.API/Dockerfile
hc-jordenlowe: i see the update below

Fabric.Authorization.API/Dockerfile
hc-jordenlowe: seems like this would open us up to using packages that do not have secure signatures.  In this article:  https://unix.stackexchange.com/questions/2544/how-to-work-around-release-file-expired-problem-on-a-local-mirror  it mentions this is a work around and instead you should use another mirror instead.  Perhaps even upgrade to a new distro for the image behind the scenes.

PR #504, opened 2019-04-17
use domain of user for querying

hc-jordenlowe:
2 scenarios:

1. This scenario.  We get the user and split it by its subject id to get Username and Domain.  Do a search and map this to AzureAD.  There were redundant calls that were removed.  Created a function to give the domain part of the name too.

2. The previous code took in a domain name.  Do we want to search AD just for the domain we have from the currently logged in user?  If we are scanning the database it makes sense to pull all users regardless of domain and map them based on that.  We will want to see how this may work.

Need some thoughts to get my head around this.

### DIFF
--- a/Fabric.Authorization.API/scripts/AzureAD-Users-Migration.ps1
+++ b/Fabric.Authorization.API/scripts/AzureAD-Users-Migration.ps1
@@ -365,9 +365,7 @@ function Get-ADUsers
 {
     param(
         [Parameter(Mandatory=$true)]
-        [System.Data.DataSet] $authDataSet,
-        [Parameter(Mandatory=$true)]
-        [string] $domain
+        [System.Data.DataSet] $authDataSet
     )
 
     # Add sourceAnchor to existing auth table (userTable) passed in
@@ -376,19 +374,27 @@ function Get-ADUsers
     {
        # cleanup variables 
        # if AD User is mispelled the values from the previous user will be entered for the next user
-       $pc = $null
        $samAccountName = $null
        $entry = $null
        $userMap = $null
        $userMapGuid = $null
 
        [string] $subjectId = $user.SubjectId.ToString()
-       # find the samAccountName in AD from the AuthorizationDB SubjectId
-       $pc = Get-PrincipalContext -domain $domain
-       $samAccountName = Get-SamAccountFromAccountName -accountName $subjectId
-
+       # This script is in Install-Authorization-Utilities.psm1
+       $samAccountName = Get-SamAccountFromAccountName -accountName $subjectId 
+       $samDomainName = Get-SamDomainFromAccountName -accountName $subjectId
+       
        # get all properties of the user with $samAccountName
-       $entry = Get-ADUser $samAccountName -Properties *
+       # You can find this script in ActiveDirectory module, located here:
+       # Remote Server Administration Tools https://www.microsoft.com/en-us/download/details.aspx?id=45520
+       try 
+       {
+            $entry = Get-ADUser $samAccountName -Server $samDomainName -Properties * 
+       } 
+       catch [Microsoft.ActiveDirectory.Management.ADServerDownException]
+       {
+            Write-DosMessage -Level "Error" -Message "Could not connect to AD Server '$($samDomainName)' for user '$($samAccountName)'."
+       }
 
        # returns a System.Byte[], needs to be converted to a base 64 string to compare with AzureAD ImmutableId         
        # the system.Guid(, ) with the comma is interpreted by powershell as passing a literal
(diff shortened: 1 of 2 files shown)


johnrulon:
I'm not entirely familiar with the code and how we query for users or groups right now, but the functionality we want from the Access Control UI is this: "a user types 'john' into the search. Access Control UI calls to IDPSS and IDPSS should return all users that match 'john' in the Windows AD and any Azure AD tenant configured with IDPSS. Does that help answer the questions? If not, let's keep discussing until we get there.

johnrulon:
and to take it another step further, it would be nice if we query those in parallel and return the results as they come in. Doing a search that takes 5 seconds is ok, but 10+ seconds is where it starts to be painful as an end user.

Fabric.Authorization.API/scripts/AzureAD-Users-Migration.ps1
```diff
@@ -376,19 +374,20 @@ function Get-ADUsers
...
        $userMapGuid = $null
 
        [string] $subjectId = $user.SubjectId.ToString()
-       # find the samAccountName in AD from the AuthorizationDB SubjectId
-       $pc = Get-PrincipalContext -domain $domain
-       $samAccountName = Get-SamAccountFromAccountName -accountName $subjectId
-
+       # This script is in Install-Authorization-Utilities.psm1
+       $samAccountName = Get-SamAccountFromAccountName -accountName $subjectId 
+       $samDomainName = SamDomainFromAccountName -accountName $subjectId
```
bsmithvt: Are you missing the `Get-` prefix?

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -398,6 +398,21 @@ function Get-SamAccountFromAccountName
...
+    param(
+        [Parameter(Mandatory=$true)]
+        [string] $accountName
+    )
+    $accountNameParts = $accountName.Split('\')
+    if ($accountNameParts.Count -ne 2) {
+        Write-DosMessage -Level "Error" -Message "Please enter an account in the form DOMAIN\account. Halting installation." 
+        throw
+    }
+    $samAccountName = $accountNameParts[0]
```
hckenmiller: maybe change to samDomainName, since this returns the domain

hc-jordenlowe: good call

Fabric.Authorization.API/scripts/AzureAD-Users-Migration.ps1
hc-jordenlowe: yep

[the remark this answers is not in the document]
hc-jordenlowe: yep

Fabric.Authorization.API/scripts/AzureAD-Users-Migration.ps1
```diff
@@ -376,19 +374,20 @@ function Get-ADUsers
...
-
+       # This script is in Install-Authorization-Utilities.psm1
+       $samAccountName = Get-SamAccountFromAccountName -accountName $subjectId 
+       $samDomainName = Get-SamDomainFromAccountName -accountName $subjectId
+       
        # get all properties of the user with $samAccountName
-       $entry = Get-ADUser $samAccountName -Properties *
+       # You can find this script in ActiveDirectory module, located here:
+       # Remote Server Administration Tools https://www.microsoft.com/en-us/download/details.aspx?id=45520
+       $entry = Get-ADUser $samAccountName -Server $samDomainName -Properties *
```
ctediz: hopefully nothing too crazy happens when try to search a domain it can't talk to

Fabric.Authorization.API/scripts/Install-Authorization-Utilities.psm1
```diff
@@ -398,6 +398,21 @@ function Get-SamAccountFromAccountName
...
+
+function Get-SamDomainFromAccountName 
+{
+    param(
+        [Parameter(Mandatory=$true)]
+        [string] $accountName
+    )
+    $accountNameParts = $accountName.Split('\')
+    if ($accountNameParts.Count -ne 2) {
+        Write-DosMessage -Level "Error" -Message "Please enter an account in the form DOMAIN\account. Halting installation."
```
ctediz: Maybe add the account name that doesn't fit the pattern expected

hc-jordenlowe:
> and to take it another step further, it would be nice if we query those in parallel and return the results as they come in. Doing a search that takes 5 seconds is ok, but 10+ seconds is painful as an end user.

This is separate.  This is the migration script, used once, at the beginning of the update to AAD.  We are querying Authorization's DB, then looping through each user and adding them to the corresponding user in AAD.  The previous code did not take into account the user's domain, so Dev.Test could be from the 'Mike tenant' or 'hqcatalyst' and this would map incorrectly to the first domain it was searching, without filtering for domain.

PR #505, opened 2019-04-18
Added Tests and printed better message for invalid account names.

hc-jordenlowe:
@ctediz I missed your comment about the subjectId having a better output.  I have added a try catch to see if we can skip those.

Also added tests for the Get-SamAccountFromAccountName function

### DIFF
--- a/Fabric.Authorization.API/scripts/Install-Authorization-Utilities.tests.ps1
+++ b/Fabric.Authorization.API/scripts/Install-Authorization-Utilities.tests.ps1
@@ -381,6 +381,56 @@ Describe 'Get-SamAccountFromAccountName tests' -Tag 'Unit'{
 
             #Act
             {Get-SamAccountFromAccountName -accountName $accountName } |Should -Throw
+        }      
+        It 'Should throw an exception for only a user name' {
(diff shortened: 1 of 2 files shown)


Fabric.Authorization.API/scripts/Install-Authorization-Utilities.tests.ps1
```diff
@@ -385,6 +385,42 @@ Describe 'Get-SamAccountFromAccountName tests' -Tag 'Unit'{
...
+            {Get-SamDomainFromAccountName -accountName $accountName } |Should -Throw
+        }
+        It 'Should throw an exception for a null account domain'{
+            #Arrange
+            $accountName = $null
+
+            #Act
+            {Get-SamDomainFromAccountName -accountName $accountName } |Should -Throw
+        }
+        It 'Should throw an exception for a empty account name'{
```
miketabet: One more potential test case - account name without the domain should also throw.

hc-jordenlowe: Added that

PR #507, opened 2019-04-22
fixed bug where app grain securable permissions were not resolved cor…

bsmithvt:
…rectly from access control UI

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
+++ b/Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
@@ -8,26 +8,27 @@ import { Observable, of } from 'rxjs';
   providedIn: 'root'
 })
 export class CurrentUserService {
-  private permissions: Array<string> = [];
-  private lastUpdateTimestamp = 0;
+  private permissionMap: Map<string, Array<string>> = new Map<string, Array<string>>();
+  private lastUpdateTimestampMap: Map<string, number> = new Map<string, number>();
 
   constructor(private authUserService: FabricAuthUserService) {
   }
 
-  getPermissions(): Observable<string[]> {
+  getPermissions(securableItem: string): Observable<string[]> {
     const currentTimeStamp = new Date().getTime();
-    if (this.lastUpdateTimestamp === 0 || (Math.abs(currentTimeStamp - this.lastUpdateTimestamp) / 60000) > 2) {
-      this.lastUpdateTimestamp = currentTimeStamp;
-      return this.authUserService.getCurrentUserPermissions().pipe(tap(userPermissionResponse => {
-        this.permissions = userPermissionResponse.permissions;
+    if (!this.lastUpdateTimestampMap.has(securableItem)
+        || this.lastUpdateTimestampMap.get(securableItem) === 0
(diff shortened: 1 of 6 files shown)


Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
```diff
@@ -8,26 +8,27 @@ import { Observable, of } from 'rxjs';
...
     const currentTimeStamp = new Date().getTime();
-    if (this.lastUpdateTimestamp === 0 || (Math.abs(currentTimeStamp - this.lastUpdateTimestamp) / 60000) > 2) {
-      this.lastUpdateTimestamp = currentTimeStamp;
-      return this.authUserService.getCurrentUserPermissions().pipe(tap(userPermissionResponse => {
-        this.permissions = userPermissionResponse.permissions;
+    if (!this.lastUpdateTimestampMap.has(securableItem)
+        || this.lastUpdateTimestampMap.get(securableItem) === 0
+        || (Math.abs(currentTimeStamp - this.lastUpdateTimestampMap.get(securableItem)) / 60000) > 2) {
+          this.lastUpdateTimestampMap.set(securableItem, currentTimeStamp);
+          return this.authUserService.getCurrentUserPermissions(securableItem).pipe(tap(userPermissionResponse => {
```
ctediz: overall, getting user permissions for a specified securable item/caching per securable right?

What was the issue before? Was access being restricted based off incorrect securables?

bsmithvt: @ctediz, correct. Rusty discovered an issue where app-grain securables could not be managed even if the user was tied to a role that had the `manageauthorization` permission for the selected app-grain securable. The `/user/permissions` route will set the securable item to the top-level securable owned by the `ClientId` in the access token if it isn't specified in the request, which in this case is `fabric-access-control` (see `SetDefaultRequest` in `UsersModule`). This prevents any permissions for the selected securable from being returned in the response.

Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
```diff
@@ -8,26 +8,27 @@ import { Observable, of } from 'rxjs';
...
   }
 
-  getPermissions(): Observable<string[]> {
+  getPermissions(securableItem: string): Observable<string[]> {
     const currentTimeStamp = new Date().getTime();
-    if (this.lastUpdateTimestamp === 0 || (Math.abs(currentTimeStamp - this.lastUpdateTimestamp) / 60000) > 2) {
-      this.lastUpdateTimestamp = currentTimeStamp;
-      return this.authUserService.getCurrentUserPermissions().pipe(tap(userPermissionResponse => {
-        this.permissions = userPermissionResponse.permissions;
+    if (!this.lastUpdateTimestampMap.has(securableItem)
```
hc-jordenlowe: This is starting to get complex.  Maybe throwing this into a new function would be good.  Perhaps this logic is an object?

Just a thought

bsmithvt: Good call - I just incorporated that suggestion.

PR #508, opened 2019-04-24
Upgrade to .net core 2.1.10

michaelvidal24:
Upgrading Fabric.Authorization to .net core 2.1.10.

To build and run with this version ensure you have these versions of the SDK and hosting components:
- dotnet core sdk 2.1.603: https://dotnet.microsoft.com/download/thank-you/dotnet-sdk-2.1.603-windows-x64-installer
- runtime and hosting bundle: https://dotnet.microsoft.com/download/thank-you/dotnet-runtime-2.1.10-windows-hosting-bundle-installer

### DIFF
--- a/Fabric.Authorization.API/Dockerfile
+++ b/Fabric.Authorization.API/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/core/aspnet:2.2
+FROM mcr.microsoft.com/dotnet/core/aspnet:2.1
 ARG source
 WORKDIR /app
 EXPOSE 5004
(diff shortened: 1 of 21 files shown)


Fabric.Authorization.API/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM microsoft/aspnetcore:1.1.2
+FROM mcr.microsoft.com/dotnet/core/aspnet:2.2
```
hc-jordenlowe: dont know if it matters, but looks like 2.1 is LTS but 2.2 is current.  Do we want to use the 2.1 release instead?  here is what I saw this on:  https://hub.docker.com/_/microsoft-dotnet-core-aspnet

michaelvidal24: It's a good point, we could back down to .net core 2.1.x, this will give us a longer runway to upgrade to 3.x.

michaelvidal24:
@miketabet - we already run multiple .net core versions side by side on our web machines to support other apps that are on 2.x. I believe the dotnet cli is smart enough to know what framework an app targets, and as long as the correct runtime is installed, will choose the appropriate runtime.

PR #509, opened 2019-04-25
fixed access control bug where Save button was incorrectly enabled on…

bsmithvt:
… Custom Group screen if user did not have appropriate permissions

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
@@ -29,7 +29,7 @@ import { User } from 'oidc-client';
   styleUrls: ['./custom-group.component.scss', '../access-control.scss']
 })
 export class CustomGroupComponent implements OnInit, OnDestroy {
-  public roles: Array<IRole> = [];
+  public rolesForGrainAndSecurable: Array<IRole> = [];
   public principals: Array<IFabricPrincipal> = [];
   public customGroups: Array<IGroup> = [];
   public associatedUsers: Array<IUser> = [];
@@ -62,7 +62,8 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
 
   private userType = 'user';
   private groupType = 'group';
-  private groupRoles: Array<IRole> = [];
+  private groupRolesForGrainAndSecurable: Array<IRole> = [];
+  private allGroupRoles: Array<IRole> = [];
   private missingPermissions: Array<string> = [];
 
   constructor(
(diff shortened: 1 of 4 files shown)


bsmithvt:
My only question around these changes is where I bypass the `Intersect` logic in the group service if the client is `fabric-access-control`. That's the only way I could get all the roles for a particular group across grains and securables to be returned. Will this change cause any issues elsewhere? I don't think it will but need input from others.

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
```diff
@@ -280,37 +247,28 @@ export class CustomGroupComponent implements OnInit, OnDestroy {
...
+  getGroup(): Observable<IGroup> {
+     return this.groupService.getGroup(this.groupName);
   }
 
   getGroupRolesBySecurableItemAndGrain(): Observable<IRole[]> {
-    return observableZip(this.getRolesBySecurableItemAndGrain(), this.getGroupRoles()).pipe(
-      map((result: [IRole[], IRole[]]) => {
-        let allRoles = result[0];
-        this.groupRoles = result[1];
+    return observableZip(this.roleService.getRolesBySecurableItemAndGrain(this.grain, this.securableItem), this.getGroup()).pipe(
```
hc-jordenlowe: From above this is getting all groups with that groupName, then you filter down based on grain and securableItem.  Wouldnt it be 1 item already?  

But thats probably needed because of how everything works in lists with observables.  if you have something left over then you have permissions, but otherwise reject it.

hc-jordenlowe:
Hey @bsmithvt where is the intersect logic?  I didnt see that operation explicitly so I might need help pin pointing it to get a clearer perspective on the issue.

Also, in reading the code, and javascript in general, I feel like smaller functions and composition is needed for Access-Control.    Maybe its me not working in it everyday but the code looks too complex to take a quick glance at and understand the lambdas after lambdas.  Food for thought on the PR.

Fabric.Authorization.Domain/Services/GroupService.cs
```diff
@@ -65,9 +66,13 @@ public async Task<Group> UpdateGroup(Group group)
 
         public async Task<Group> GetGroup(GroupIdentifier groupIdentifier, string clientId)
         {
-            var group = await _groupStore.Get(groupIdentifier);             
-            var clientRoles = (await _roleService.GetRoles(clientId)).ToList();
-            group.Roles = clientRoles.Intersect(group.Roles).ToList();
+            var group = await _groupStore.Get(groupIdentifier);
+            if (!string.Equals(clientId, AccessControl.ClientId, StringComparison.OrdinalIgnoreCase))
```
hc-jordenlowe: We are saying that if we are looking at any other group besides "Fabric-Access-Control", take the groups roles and only return the ones that are based on the Client's roles and the Groups roles.  

So the difference was that we were doing this, even for "Fabric-Access-Control".  Now when you get the group for "Fabric-Access-Control", it will have all roles.  Is that effectively saying that any role associated outside "Fabric-Access-Control" can edit it?

Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.ts
bsmithvt: Which method are you referring to? `getGroup` will always return a single group. I'm filter the roles by grain and securableItem.

bsmithvt:
> Hey @bsmithvt where is the intersect logic? Edit: I see it now. its at the very bottom
> 
> Also, in reading the code, and javascript in general, I feel like smaller functions and composition is needed for Access-Control. Maybe its me not working in it everyday but the code looks too complex to take a quick glance at and understand the lambdas after lambdas. Food for thought on the PR.

Fabric.Authorization.Domain/Services/GroupService.cs
bsmithvt: No. This is just allowing the `fabric-access-control` client to retrieve roles that are not "owned" by it (see the `ClientOwner` field on the `SecurableItems` table. Users still need to have certain permissions to edit groups tied to a particular grain/securable item.

bsmithvt:
@hc-jordenlowe - the intersect logic is located in `GroupService.cs`, line 73.

PR #513, opened 2019-05-01
Roles getting unchecked if group exists

joshnielsen10:
If the user tries to create new custom group with a group name that already exists, the roles would get unchecked when they changed the group name.  So I added an if statement to see if the user got this error, and if so, then the roles wouldn't get unchecked.

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.spec.ts
+++ b/Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.spec.ts
@@ -263,6 +263,36 @@ describe('CustomGroupComponent', () => {
         + `${mockUsersResponse[0].identityProviderUserPrincipalName}`);
     });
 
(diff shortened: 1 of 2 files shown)


Fabric.Authorization.AccessControl/src/app/modules/access-control/custom-group/custom-group.component.spec.ts
```diff
@@ -263,6 +263,36 @@ describe('CustomGroupComponent', () => {
         + `${mockUsersResponse[0].identityProviderUserPrincipalName}`);
     });
 
+    it('dont reset the roles if group name invalid', () => {
```
hc-jordenlowe: YOU DID IT!  Great job!  Looks like we werent too far off.  Thats awesome!

joshnielsen10: Thanks to you and your help, I appreciate it.

PR #520, opened 2019-05-23
'bandaid' for race condition

ctediz:
There is a race condition occurring, due to the UI tests speeding through the initial member-list page which hits the current-user service.  Then the test navigates to the member page which also hits the user service. 

Occasionally, the api call to authorization would not yet be resolved so the UI test would get back the cached permissions which were still undefined.

### DIFF
--- a/Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
+++ b/Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
@@ -1,5 +1,5 @@
 
-import {map, tap} from 'rxjs/operators';
+import { map, tap } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { FabricAuthUserService } from './fabric-auth-user.service';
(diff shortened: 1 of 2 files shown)


Fabric.Authorization.AccessControl/src/app/services/current-user.service.ts
```diff
@@ -17,10 +17,13 @@ export class CurrentUserService {
   getPermissions(securableItem: string): Observable<string[]> {
     const currentTimeStamp = new Date().getTime();
     if (this.shouldRefreshCache(securableItem, currentTimeStamp)) {
+      return this.authUserService.getCurrentUserPermissions(securableItem).pipe(
+         tap(userPermissionResponse => {
           this.lastUpdateTimestampMap.set(securableItem, currentTimeStamp);
```
ctediz: Moved this section into the observable, so the cache should hopefully not be returned until a value is returned.

PR #540, opened 2019-11-12
Remove idPSS discovery check on startup + other uses of idPSS settings

amunar:
- Removed the check on startup for idPSS endpoint settings from discovery service.  This check was causing a 500 error during Auth install or API startup if an idPSS was not registered with the discovery service.
- Removed code related to calling to discovery service (the idPSS check was the last use).
- Moved method for searching groups from idPSS to the Identity service (was the last code still referencing the idPSS endpoint setting).
- Removed namespace for idPSS service.
- Removed idPSS settings from appsettings.json and IAppConfiguration.


Fabric.Authorization.API/RemoteServices/Identity/Providers/IdentityServiceProvider.cs
```diff
@@ -25,6 +25,8 @@ public class IdentityServiceProvider : IIdentityServiceProvider
         private readonly ILogger _logger;
 
         private const string ServiceName = "Fabric.Identity";
+        private const string UsersEndpoint = "users";
+        private const string PrincipalsEndpoint = "principals/";
```
ctediz: should these both contain, or not contain slashes?

amunar: UsersEndpoint can have a slash for more consistency.

PR #543, opened 2019-12-17
Disable Windows authentication during install

amunar:
Install scripts modified to force Windows authentication to be disabled for Authorization.

Windows Authentication for Authorization needs to be disabled to pass later install checks.
Running install scripts would inherit the setting from the web site Authorization is installed under, potentially leaving it enabled.

Relevant work item: 192656


hckenmiller:
If we set windowsAuthentication enabled="false" in the web.config, would this accomplish the same thing. Do we know how windowsAuthentication is getting set to enabled.

amunar:
> If we set windowsAuthentication enabled="false" in the web.config, would this accomplish the same thing. Do we know how windowsAuthentication is getting set to enabled.

Yup, the Set-WebConfigurationProperty is setting what's shown in the filter option in its web.config, with the name option being the attribute being set.

Before this change, the Windows authentication setting was being inherited by the authentication setting on the site it was being installed under.  If the web site's Windows authentication was set to enabled, Authorization's setting was also enabled.

